PR #7, opened 2023-10-20
Dpr2 148 caseload

gavriil-g-moj:
Step 0 of using caseload ids for row level security. 
Note: The ConfiguredApiIntegrationTest fails due to auth issues within the test context. The end to end integration test has been performed in the mi appication.
In the future we might need to create a separate small application inside the library for integration testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,12 @@ You will also need to add the following to your Spring Boot application class:
 `@ComponentScan("yourapplicationpackage","uk.gov.justice.digital.hmpps.digitalprisonreportinglib")`
 Where you will need to replace "yourapplicationpackage" with the actual package of your application.
 
+You will also need to instantiate the AuthAwareAuthenticationToken class in order to pass as a 3rd parameter the list of active caseload ids since this is used
+for row level security in the library, as the JWT token is needed to retrieve the caseload user details.
+Retrieving these details can be done simply by calling the getActiveCaseloadIds method of the CaseloadService and passing the JWT as a parameter.
+An example can be found [here](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-mi/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportingmi/security/AuthAwareTokenConverter.kt#L15).
+The AuthAwareAuthenticationToken can then be used in your controllers to retrieve the caseload ids should you need to implement row level security. 
+
 ### Open API Docs
 The API documentation is generated via the following dependency:
 `implementation("org.springdoc:springdoc-openapi-starter-webmvc-ui:2.2.0")`
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -31,6 +31,7 @@ dependencies {
   // Testing
   testImplementation("com.h2database:h2")
   testImplementation("javax.xml.bind:jaxb-api:2.4.0-b180830.0359")
+  testImplementation("io.jsonwebtoken:jjwt:0.12.2")
 }
 
 java {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -22,6 +22,8 @@ dependencies {
   implementation("org.springframework.boot:spring-boot-starter-data-jpa")
 
   implementation("com.google.code.gson:gson:2.8.5")
+  implementation("org.springframework.boot:spring-boot-starter-security")
+  implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
 
   // Swagger
   implementation("org.springdoc:springdoc-openapi-starter-webmvc-ui:2.2.0")
@@ -67,7 +69,7 @@ publishing {
         group = "uk.gov.justice.service.hmpps"
         name.set(base.archivesName)
         artifactId = base.archivesName.get()
-        version = "1.0.0-rc.2"
+        version = "1.0.0-rc.3-SNAPSHOT"
         description.set("A Spring Boot reporting library to be integrated into your project and allow you to produce reports.")
         url.set("https://github.com/ministryofjustice/hmpps-digital-prison-reporting-mi")
         licenses {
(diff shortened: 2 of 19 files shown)


README.md
```diff
@@ -89,6 +89,12 @@ You will also need to add the following to your Spring Boot application class:
 `@ComponentScan("yourapplicationpackage","uk.gov.justice.digital.hmpps.digitalprisonreportinglib")`
 Where you will need to replace "yourapplicationpackage" with the actual package of your application.
 
+You will also need to instantiate the AuthAwareAuthenticationToken class in order to pass as a 3rd parameter the list of active caseload ids since this is used
+for row level security in the library, as the JWT token is needed to retrieve the caseload user details.
+Retrieving these details can be done simply by calling the getActiveCaseloadIds method of the CaseloadService and passing the JWT as a parameter.
+An example can be found [here](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-mi/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportingmi/security/AuthAwareTokenConverter.kt#L15).
+The AuthAwareAuthenticationToken can then be used in your controllers to retrieve the caseload ids should you need to implement row level security.
```
stalwartsolutions, line 96: I think "instantiate" is a bit vague. Maybe you say it needs to be used or extended in their Spring Security configuration as the user's Authentication implementation. Or something similar.

build.gradle.kts
```diff
@@ -67,7 +70,7 @@ publishing {
         group = "uk.gov.justice.service.hmpps"
         name.set(base.archivesName)
         artifactId = base.archivesName.get()
-        version = "1.0.0-rc.2"
+        version = "1.0.0-rc.3-SNAPSHOT"
```
stalwartsolutions, line 73: Should this still be a snapshot?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ConfiguredApiController.kt
```diff
@@ -53,8 +55,18 @@ class ConfiguredApiController(val configuredApiService: ConfiguredApiService) {
...
-    return configuredApiService.validateAndFetchData(reportId, reportVariantId, filtersOnly(filters), selectedPage, pageSize, sortColumn, sortedAsc)
+    return configuredApiService.validateAndFetchData(
+      reportId,
+      reportVariantId,
+      filtersOnly(filters),
+      selectedPage,
+      pageSize,
+      sortColumn,
+      sortedAsc,
+      (authentication as AuthAwareAuthenticationToken).caseloads,
```
stalwartsolutions, line 68: Do you have to cast here? Can you ask it for a `AuthAwareAuthenticationToken` rather than an `Authentication`  in the controller method? Or does it only recognise Authentication?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
```diff
@@ -29,16 +29,16 @@ class ConfiguredApiRepository {
     pageSize: Long,
     sortColumn: String,
     sortedAsc: Boolean,
+    caseloads: List<String>,
   ): List<Map<String, Any>> {
-    val (preparedStatementNamedParams, whereClause) = buildWhereClause(filtersExcludingRange, rangeFilters)
+    if (caseloads.isEmpty()) {
```
stalwartsolutions, line 34: I think this should log a warning, or possibly an error. Also, it's possible that a specific query doesn't need to be filtered by caseload.

Maybe `buildWhereClause` could throw an unauthorised exception if it expected a caseload, but none were provided?

build.gradle.kts
gavriil-g-moj: Good spot.
No, that was just for my local testing. 
This will be changed to rc-3.

src/test/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/integration/ConfiguredApiIntegrationTest.kt
```diff
@@ -30,6 +31,9 @@ import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.PrisonerEntit
 import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.PrisonerRepository
 import java.time.LocalDateTime
 
+//This currently fails as the auth set up is not working correctly in test.
+//The integration test has been performed in the mi application with the new version of the library
+@Disabled
```
stalwartsolutions, line 36: This makes me a bit uneasy. If we extend the library in future for another client, the coverage is in a repo that may not be updated yet. Is there a way to disable the security for the tests, if it's too much of an obstacle?

README.md
gavriil-g-moj: Makes sense, I have rephrased  this now.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ConfiguredApiController.kt
gavriil-g-moj: Good point, initially I was under the impression the AuthAwareAuthenticationToken could not be passed directly but it seems it is working fine so I have now changed this.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
gavriil-g-moj: I returned an empty list based on the Jira AC that if a user is authenticated and has no caseload they see 0 records and the count returned is 0.
I am not sure if we want to throw an unauthorised exception since the row level security acts as a filter and if nothing matches simply no rows are returned. 
Similarly there could also be another case for example when the user has one or more caseloads and nothing matches and again they see zero results. 
Would we want to treat this also as unauthorised? 
Happy to discuss to clarify this.

src/test/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/integration/ConfiguredApiIntegrationTest.kt
gavriil-g-moj: I have fixed the integration tests now and they are in the library .

stalwartsolutions (approved):

PR #10, opened 2023-10-25
[DPR-122] Externalisable product definition

stalwartsolutions:
This PR changes the Product Definition Repository and its date converter into default implementations of interfaces, which are autoconfigured in the absence of other beans that fulfil the interface.

This means that the client application should be able to provide its own implementation (e.g. sourced from DDB, hardcoded in the app) while leaving the existing API usage of the library's definitions unchanged.

Alternatively, the client can set the property `lib.definition.location` to use the default repository implementation with their own JSON production definition.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
@@ -12,7 +12,6 @@ import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.controller.Configu
 import java.sql.Timestamp
 
 @Service
-@Suppress("UNCHECKED_CAST")
 class ConfiguredApiRepository {
 
   companion object {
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
```diff
@@ -78,26 +79,32 @@ class ConfiguredApiRepository {
...
-    val caseloadsStringArray = "(${caseloads.map { "\'$it\'" }.joinToString()})"
-    val caseloadsWhereClause = "origin_code IN $caseloadsStringArray OR  destination_code IN $caseloadsStringArray"
+    val caseloadsWhereClause = buildCaseloadsWhereClause(caseloads, caseloadFields)
     val whereClause = allFilters?.let { "WHERE $it AND ($caseloadsWhereClause)" } ?: "WHERE $caseloadsWhereClause"
     return Pair(preparedStatementNamedParams, whereClause)
   }
 
+  private fun buildCaseloadsWhereClause(caseloads: List<String>, caseloadFields: List<String>): String {
+    if (caseloadFields.isEmpty()) {
+      return "TRUE"
```
gavriil-g-moj, line 96: This is an interesting scenario because if the report definition contains no caseload fields we are not interested in  checking the user caseloads. 
But for the scenario when the user has no caseloads it will return and empty list regardless of no caseload fields.
But for phase 0 this is fine, we will implement this in the next phase.

gavriil-g-moj (approved):
LGTM

PR #13, opened 2023-10-25
[DPR-122] Refactor auth and caseload handling for client QoL

stalwartsolutions:


build.gradle.kts
```diff
@@ -72,9 +72,9 @@ publishing {
         group = "uk.gov.justice.service.hmpps"
         name.set(base.archivesName)
         artifactId = base.archivesName.get()
-        version = "1.0.0-rc.6"
+        version = "1.0.0-rc.7"
         description.set("A Spring Boot reporting library to be integrated into your project and allow you to produce reports.")
-        url.set("https://github.com/ministryofjustice/hmpps-digital-prison-reporting-mi")
+        url.set("https://github.com/ministryofjustice/hmpps-digital-prison-reporting-mi-lib")
```
gavriil-g-moj, line 77: Good spot, this should be `https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib` without the 'mi' part.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/ResourceServerConfiguration.kt
```diff
@@ -1,34 +1,46 @@
...
   @Throws(Exception::class)
-  fun filterChain(http: HttpSecurity, caseloadService: CaseloadService): SecurityFilterChain? {
+  fun filterChain(http: HttpSecurity, caseloadProvider: CaseloadProvider, tokenConverter: AuthAwareTokenConverter): SecurityFilterChain? {
     http {
       headers { frameOptions { sameOrigin = true } }
       sessionManagement { sessionCreationPolicy = SessionCreationPolicy.STATELESS }
       // Can't have CSRF protection as requires session
       csrf { disable() }
       authorizeHttpRequests {
+        listOf(
```
gavriil-g-moj, line 27: This class is for integration tests only so there is no need to whitelist the health and swagger endpoints.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/ResourceServerConfiguration.kt
```diff
@@ -1,34 +1,46 @@
...
 import org.springframework.security.web.SecurityFilterChain
-import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.service.CaseloadService
 
 @Configuration
-class ResourceServerConfiguration {
-
-  @Value("\${spring.security.user.roles}")
-  private lateinit var authorisedRole: String
+@ConditionalOnProperty(name = ["dpr.lib.user.role"])
+class ResourceServerConfiguration(
```
gavriil-g-moj, line 14: By default Spring Boot will return unauthorised for requests when the spring security dependency is added through Gradle.
If the dpr.lib.user.role property is not defined then this bean will not be created and all requests will result in 401.
Would this be the intention of this conditional property?

stalwartsolutions: Yes, so that's also the current behaviour - this library will introduce the security dependency, blocking all requests until the client app adds security config.

This gives them a default implementation to get around that without writing their own. :)

gavriil-g-moj (approved):
LGTM

PR #15, opened 2023-10-27
[DPR-122] Swagger and config tweaks

stalwartsolutions:


build.gradle.kts
```diff
@@ -72,7 +72,7 @@ publishing {
         group = "uk.gov.justice.service.hmpps"
         name.set(base.archivesName)
         artifactId = base.archivesName.get()
-        version = "1.0.0-rc.8"
+        version = "1.0.1"
```
gavriil-g-moj, line 75: Should this be 1.0.0 instead? We haven't had 1.0.0 released yet. All of the previous versions were release candidates.

gavriil-g-moj (approved):
It looks good to me. 
Only one comment regarding changing the published version to 1.0.0

PR #20, opened 2023-10-31
[DPR2-189] Tweak for date-time filtering

stalwartsolutions:
- Fixed issue with date-times being filtered excluding the last day of the filter.
- Refactored data repo so that it accepts filters, with metadata about what type of filter they are.
- Added methods to get simplified product/report definition if only a single report is being worked with.
- Changed data service to validate and map to new filters with metadata, rather than just validate.


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
```diff
@@ -104,7 +104,7 @@ class ConfiguredApiRepository {
       if (k.endsWith(RANGE_FILTER_START_SUFFIX)) {
         "${k.removeSuffix(RANGE_FILTER_START_SUFFIX)} >= :$k"
       } else if (k.endsWith(RANGE_FILTER_END_SUFFIX)) {
-        "${k.removeSuffix(RANGE_FILTER_END_SUFFIX)} <= :$k"
+        "${k.removeSuffix(RANGE_FILTER_END_SUFFIX)} < DATEADD(day, 1, CAST(:$k AS timestamp))"
```
gavriil-g-moj, line 107: This will only work for the range filter being a date, but the range filters might be other types in the future. 
Shall we have a separate condition instead for the date.end case?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ProductDefinitionRepositoryAutoConfig.kt
```diff
@@ -7,7 +7,7 @@ import org.springframework.context.annotation.Configuration
 
 @Configuration
 class ProductDefinitionRepositoryAutoConfig(
-  @Value("\${dpr.lib.definition.location:productDefinition.json}") private val definitionResourceLocation: String,
+  @Value("\${dpr.lib.definition.location}") private val definitionResourceLocation: String,
```
gavriil-g-moj, line 10: If this gets updated shall we also update the readme to reflect that the `productDefinition.json` will not be the default value.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
stalwartsolutions: Well that's a good point! I hadn't considered that this endpoint could support other types of range in future. I'll have a little refactor to see if I can cater for both scenarios.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ProductDefinitionRepositoryAutoConfig.kt
stalwartsolutions: Actually, it would probably be best to leave this as it is, thinking about it. Perhaps with an empty definition as a placeholder.

src/test/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepositoryTest.kt
```diff
@@ -474,7 +463,7 @@ class ConfiguredApiRepositoryTest {
       4,
       7849,
       LocalDateTime.of(2023, 5, 1, 0, 0, 0),
-      LocalDateTime.of(2023, 5, 1, 15, 19, 0),
+      LocalDateTime.of(2023, 5, 1, 1, 15, 19),
```
gavriil-g-moj, line 466: Is this change required?

stalwartsolutions: Ah, that would be a mistake. I'll take that out.

gavriil-g-moj:
It looks good to me. 
My main comment would be if we can have a test with a case that passes with the code change that adds one day to the end date but would fail if you simply had the less than or equal to comparison as it was before.

gavriil-g-moj (approved):
LGTM

PR #28, opened 2023-11-13
[DPR-000] Load caseloads lazily

stalwartsolutions:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/AuthAwareAuthenticationToken.kt
```diff
@@ -5,12 +5,23 @@ import org.springframework.security.oauth2.jwt.Jwt
...
 class AuthAwareAuthenticationToken(
-  jwt: Jwt,
+  val jwt: Jwt,
   private val aPrincipal: String,
   authorities: Collection<GrantedAuthority>,
-  val caseloads: List<String>,
+  private val caseloadProvider: CaseloadProvider,
 ) : JwtAuthenticationToken(jwt, authorities) {
+
+  private var caseloads: List<String>? = null
```
gavriil-g-moj, line 14: I'm thinking this might need to be thread safe.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/AuthAwareAuthenticationToken.kt
```diff
@@ -5,12 +5,23 @@ import org.springframework.security.oauth2.jwt.Jwt
...
+
   override fun getPrincipal(): String {
     return aPrincipal
   }
+
+  fun getCaseLoads(): List<String> {
```
gavriil-g-moj, line 20: This method probably needs to be synchronized.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/AuthAwareAuthenticationToken.kt
gavriil-g-moj: Double checked in the code and confirmed with Stu that the AuthAwareAuthenticationToken is not shared and it is created per request.

gavriil-g-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/AuthAwareAuthenticationToken.kt
gavriil-g-moj: This might need to be synchronized.
Update: Double checked in the code and confirmed with Stu that the AuthAwareAuthenticationToken is not shared and it is created per request.

PR #31, opened 2023-11-19
Dpr2 221 align model with report builder

gavriil-g-moj:
This PR includes changes to align the product definition model with the model the report builder produces.

The main changes are:
- Changes to some of the json field names.
- A change to parse the report definition as a json object instead of an array.

There are some **questions** coming up as part of this PR and DPR2-221:
1. Should the field name changes of the data model be reflected also to the controller model? For example the date range filter type is defined as date-range in the data model while the json response the library produces defines it as "dateRange". There are similar examples for other fields that were changed to align with the report builder.
My thoughts are that they should be the same both in the data model and in the controller model for consistency. 
Also I would be in favour of changing the report builder to produce camel case rather than kebab-case. 
Google suggest this as well:
https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format
2. Currently there is a /definitions endpoint. After the changes to read from a single report definition should this endpoint be deleted? Do we still need an endpoint to return the single report definition? i.e. `/definition`

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/ReportField.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/ReportField.kt
@@ -1,10 +1,12 @@
 package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.model
 
 data class ReportField(
-  val schemaField: String,
-  val displayName: String,
+  val name: String,
+  val display: String,
   val wordWrap: WordWrap? = null,
   val filter: FilterDefinition? = null,
   val sortable: Boolean = true,
-  val defaultSortColumn: Boolean = false,
+  val `default-sort`: Boolean = false,
+  val formula: String? = null,
+  val visible: Boolean?,
 )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/model/FieldDefinition.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/model/FieldDefinition.kt
@@ -2,7 +2,7 @@ package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.controller.model
 
 data class FieldDefinition(
   val name: String,
-  val displayName: String,
+  val display: String,
   val wordWrap: WordWrap? = null,
   val filter: FilterDefinition? = null,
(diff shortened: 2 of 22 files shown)


dominic-messenger-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/ReportField.kt
```diff
@@ -1,10 +1,12 @@
...
 data class ReportField(
-  val schemaField: String,
-  val displayName: String,
+  val name: String,
+  val display: String,
   val wordWrap: WordWrap? = null,
   val filter: FilterDefinition? = null,
   val sortable: Boolean = true,
-  val defaultSortColumn: Boolean = false,
+  val `default-sort`: Boolean = false,
```
stalwartsolutions, line 9: Is it possible to have the source JSON kebab-case, without kebabing the Kotlin property? Could `@JsonProperty` or similar be used here?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/ReportField.kt
```diff
@@ -1,10 +1,12 @@
...
+  val formula: String? = null,
+  val visible: Boolean?,
```
stalwartsolutions, line 11: Is it worth noting on unused properties that they're currently not mapped pending ticket DPR2-XXX?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/ReportField.kt
gavriil-g-moj: This has now been changed to defaultsort so the kebab case is removed.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/ReportField.kt
gavriil-g-moj: I have mentioned this in a comment and created https://dsdmoj.atlassian.net/browse/DPR2-241 to link to the comment.

stalwartsolutions (approved):

stalwartsolutions (approved):

PR #45, opened 2023-12-04
[DPR-198] Data model tweaks

stalwartsolutions:


tom-ogle-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/FilterType.kt
```diff
@@ -1,6 +1,8 @@
 package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.model
 
-enum class FilterType(val type: String) {
+import com.fasterxml.jackson.annotation.JsonValue
+
+enum class FilterType(@JsonValue val type: String) {
```
gavriil-g-moj, line 5: I believe this is not needed as the FilterTypeDeserializer handles this.

PR #46, opened 2023-12-04
DPR2-72 return static options from definition endpoint when returnAsStaticOptions is true

gavriil-g-moj:
This is the second PR for DPR2-72 to fulfil the following AC:

> If returnAsStaticOptions is true:
> 
> The unique values for the field are requested when the definition is requested, and included as staticOptions (as with the existing ones).

When the report definition is requested, for each report field which is a dynamic filter and has the flag returnAsStaticOptions set to true, the list of static options is populated in the response and contains all the available distinct values of that field. 
There is no prefix filtering in this case nor any other filters when performing the SQL query, apart from the caseload checks.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/ReportDefinitionMapper.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/ReportDefinitionMapper.kt
@@ -29,34 +29,40 @@ class ReportDefinitionMapper(val configuredApiService: ConfiguredApiService) {
   val todayRegex: Regex = Regex("today\\(\\)")
   val dateRegex: Regex = Regex("today\\((-?\\d+), ?([a-z]+)\\)", RegexOption.IGNORE_CASE)
 
-  fun map(productDefinition: ProductDefinition, renderMethod: RenderMethod?, caseLoads: List<String>): ReportDefinition = ReportDefinition(
+  fun map(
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/ReportDefinitionMapper.kt
```diff
@@ -83,19 +93,42 @@ class ReportDefinitionMapper {
...
+    filterDefinition: uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.model.FilterDefinition,
+    productDefinitionId: String,
+    reportVariantId: String,
+    schemaFieldName: String,
+    caseLoads: List<String>,
+  ): List<FilterOption>? {
+    return filterDefinition.dynamicOptions?.takeIf { it.returnAsStaticOptions }?.let {
+      configuredApiService.validateAndFetchData(
+        productDefinitionId,
+        reportVariantId, emptyMap(), 1, 10, schemaFieldName, true, caseLoads, schemaFieldName,
```
stalwartsolutions, line 128: Will this only return the first 10 options?

gavriil-g-moj: Yes it is hardcoded here but we can make it configurable.

stalwartsolutions (approved):

PR #54, opened 2023-12-13
[DPR-181] Return message instead of erroring if user has no active caseload

stalwartsolutions:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ConfiguredApiController.kt
```diff
@@ -59,23 +78,47 @@ class ConfiguredApiController(val configuredApiService: ConfiguredApiService) {
...
+            reportVariantId,
+            filtersOnly(filters),
+            selectedPage,
+            pageSize,
+            sortColumn,
+            sortedAsc,
+            authentication.getCaseLoads(),
+          ),
+        )
+    } catch (exception: NoDataAvailableException) {
```
gavriil-g-moj, line 97: Would it be better to move the exception handling to the DigitalPrisonReportingExceptionHandler instead of a catch block in the controller?

stalwartsolutions: Yes, but the returned value is dependent on the endpoint - the data endpoint returns a list, but the count endpoint returns a "Count".

gavriil-g-moj (approved):
LGTM

PR #58, opened 2023-12-18
[DPR-000] Add bean prefixes and allow non-DPR auth tokens

stalwartsolutions:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/PolicyEngine.kt
```diff
@@ -35,7 +35,7 @@ class PolicyEngine(
       }
       // Note: This is currently for a single active caseload
       // Addition of single quotes could be in DPD instead
-      interpolated = s.replace(CASELOAD, "${authToken.getCaseLoads().first()}")
+      interpolated = s.replace(CASELOAD, authToken.getCaseLoads().first())
```
gavriil-g-moj, line 38: This needs to contain the ${} that define the variables as these get replaced as well.

gavriil-g-moj (approved):
LGTM

PR #61, opened 2023-12-20
[DPR-000] Allow non-DPR auth tokens to be used

stalwartsolutions:


gavriil-g-moj:
Currently we have a conditional DprAuthAwareTokenConverter bean in the TokenConverterAutoConfig which will be created if there is no existing DprAuthAwareTokenConverter.
If we will allow different types of tokens other than DPR would we need to make changes to the above as well?

PR #65, opened 2024-01-09
Dpr2 300 Configuration for reading from definitions service

gavriil-g-moj:
This PR includes changes to allow the library to have a client configured to read data product definitions from a definitions service.
- The host application can specify a dataProductDefinitions.host property and not specify the definition.locations and the library will call that endpoint to retrieve the definitions instead. The path to the definitions directory is provided as a query parameter in the request.

Pending changes:
- Add the dataProductDefinitionsPath also to ConfiguredApiController.


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ReportDefinitionController.kt
```diff
@@ -41,12 +47,19 @@ class ReportDefinitionController(val reportDefinitionService: ReportDefinitionSe
     @RequestParam("maxStaticOptions", defaultValue = "20")
     @Min(1)
     maxStaticOptions: Long,
+    @Parameter(
+      description = dataProductDefinitionsPathDescription,
+      example = dataProductDefinitionsPathExample,
+    )
+    @RequestParam("dataProductDefinitionsPath")
+    dataProductDefinitionsPath: String? = null,
```
stalwartsolutions, line 55: Please could you set these new parameters to the orphanage path by default, to save making changes in the UI?

stalwartsolutions: Also, should `/definitions` be always used as the root of the path, given that this is the definitions endpoint?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ProductDefinitionRepositoryAutoConfig.kt
```diff
@@ -3,24 +3,40 @@ package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data
...
     localDateTimeTypeAdaptor: LocalDateTimeTypeAdaptor,
     dprDefinitionGson: Gson,
   ): ProductDefinitionRepository = JsonFileProductDefinitionRepository(
-    definitionResourceLocations,
+    definitionResourceLocations ?: emptyList(),
     dprDefinitionGson,
   )
 
+  @Bean
+  @ConditionalOnProperty(prefix = "dpr.lib.definition", name = ["locations"], matchIfMissing = true)
```
stalwartsolutions, line 30: Should this be conditional on the host property being present too?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ReportDefinitionController.kt
gavriil-g-moj: I have now defaulted the new parameter to the orphanage path: `definitions /prisons/orphanage`
Regarding hardcoding `/definitions` as the root path, this is would not always be the case as this would depend on the given implementation of the service which serves the definitions.
In our case [the root of the path](https://github.com/ministryofjustice/hmpps-dpr-data-product-definitions/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/dprdpdsservice/controller/DefinitionsController.kt#L16) indeed has `/definitions` but some other implementation might choose to differ.
So if we hardcoded it, it might not work with other implementations of the definitions service.

gavriil-g-moj: Alternatively, this can be set by the host application in the dpr.lib.caseloads.host configuration just once.
i.e. `http://localhost:9999/definitions`

gavriil-g-moj: And maybe I could rename the property name to rootPath rather than host?

stalwartsolutions: Good points. I d think it's fine as it is. :)

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ProductDefinitionRepositoryAutoConfig.kt
gavriil-g-moj: I have now changed this to be created if the locations property is absent and the host property is present.

gavriil-g-moj: I just noticed some tests are still not happy with this and I am looking into it.

gavriil-g-moj: Fixed now.

stalwartsolutions (approved):

PR #69, opened 2024-01-19
[DPR-000] Named datasource spike

stalwartsolutions:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
```diff
@@ -33,7 +40,7 @@ class ConfiguredApiRepository {
     dynamicFilterFieldId: String? = null,
   ): List<Map<String, Any>> {
     val stopwatch = StopWatch.createStarted()
-    val result = jdbcTemplate.queryForList(
+    val result = JdbcTemplate(context.getBean("customDataSource", DataSource::class) as DataSource).queryForList(
```
stalwartsolutions, line 43: The DPD's datasource name could be used here.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/CustomDataSourceConfig.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import javax.sql.DataSource
+
+@Configuration
+class CustomDataSourceConfig(
+  @Value("\${spring.datasource.url}") private val url: String,
+  @Value("\${spring.datasource.username}") private val username: String,
+  @Value("\${spring.datasource.password}") private val password: String,
+  @Value("\${spring.datasource.driver-class-name}") private val driver: String,
```
stalwartsolutions, line 14: Any config values could be used here.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/CustomDataSourceConfig.kt
```diff
@@ -0,0 +1,26 @@
...
+) {
+  @Bean("customDataSource")
```
gavriil-g-moj, line 16: The problem here is that the bean name cannot be dynamic and derived from the DPD.

stalwartsolutions: This is the part that would be in the client application though. So for example in the Preview Tool we could configure our RedShift datasource, and then add other DPS datasources over time. Then the DPDs could refer to those datasources without any changes being made to the library.

gavriil-g-moj: So by taking this example here, the client application would have a custom datasource defined in its application.yml called "customDataSource" correct?
It just means that the clients would have to use the "customDataSource" predefined name.

stalwartsolutions: What do you mean by predefined? It's in their code (not the library), and in their DPD, and the two would have to match up. So it's predefined in that sense.

gavriil-g-moj: Ok that could work. Yes indeed their code and the DPD would have to have the same datasource name and match but also the library would have to know about the client datasource name. 
Which since it would be what has been defined in the DPD can be retrieved from the DPD when getting the [DataSource bean here](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/pull/69/files#diff-0ed8df46b0d441fbf71b237a7e141d6bcb4339e72e5904f521b5beaf541884ccR43).

PR #73, opened 2024-01-24
Dpr2 343 named datasource

gavriil-g-moj:
This PR allows the host application to use its own DataSource to run queries.
If the host application does not define a DataSource Bean then the default DataSource will be used with the default properties as before.
If the host application creates a custom DataSource Bean it will have to name it with the DataSource name of their product definition.


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
```diff
@@ -49,6 +55,19 @@ class ConfiguredApiRepository {
     log.debug("Query Execution time in ms: {}", stopwatch.time)
     return result
   }
+
+  private fun populateJdbcTemplate(dataSourceName: String?): NamedParameterJdbcTemplate {
+    val dataSource: DataSource = dataSourceName?.let {
+      if (context.containsBean(dataSourceName)) {
```
stalwartsolutions, line 61: If there's no bean with the configured name then I think we should at least log a warning, and possibly consider raising an exception. Better that they see the right problem to solve - otherwise they might waste time trying to find out why their table is missing, when actually it's using another datasource.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/DefaultDataSourceConfig.kt
```diff
@@ -0,0 +1,34 @@
...
+  @Value("\${spring.datasource.url}") private val url: String,
+  @Value("\${spring.datasource.username}") private val username: String,
+  @Value("\${spring.datasource.password}") private val password: String,
+  @Value("\${spring.datasource.driver-class-name}") private val driver: String,
+) {
+
+  @Autowired
+  lateinit var context: ApplicationContext
+
+  @ConditionalOnMissingBean
```
stalwartsolutions, line 24: I'm surprised this created. I guess it's because it's got a different name to the default spring data one.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
```diff
@@ -31,8 +35,10 @@ class ConfiguredApiRepository {
     reportId: String,
     policyEngineResult: String,
     dynamicFilterFieldId: String? = null,
+    dataSourceName: String? = null,
```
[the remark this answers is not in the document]
gavriil-g-moj: It can default to DEFAULT_DATASOURCE but unfortunately this would not skip the null check through let due to the fact the this is an optional parameter.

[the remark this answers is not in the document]
gavriil-g-moj: If on the other hand I don't make it optional a dataSourceName value will have to be passed on every occasion this method is called which is not really needed.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/DefaultDataSourceConfig.kt
gavriil-g-moj: Yes exactly this is the reason it is created.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
```diff
@@ -49,6 +55,19 @@ class ConfiguredApiRepository {
...
   }
+
+  private fun populateJdbcTemplate(dataSourceName: String?): NamedParameterJdbcTemplate {
+    val dataSource: DataSource = dataSourceName?.let {
+      if (context.containsBean(dataSourceName)) {
+        context.getBean(dataSourceName, DataSource::class) as DataSource
+      } else {
+        context.getBean(DEFAULT_DATASOURCE, DataSource::class) as DataSource
+      }
+    } ?: context.getBean(DEFAULT_DATASOURCE, DataSource::class) as DataSource
```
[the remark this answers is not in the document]
gavriil-g-moj: I have now done that and made some more changes to simplify things.
I have removed the explicit creation of the default DataSource as this is already created by Spring and since now we are retrieving it by class if the name is not found.
As a result I have also removed the DEFAULT_DATASOURCE String and also I have made the dataSourceName parameter required.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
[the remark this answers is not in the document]
gavriil-g-moj: I have now removed the DEFAULT_DATASOURCE String and also I have made the dataSourceName parameter required.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
gavriil-g-moj: A warning is now logged if there is no Bean with the configured name. 
An exception is not thrown as there is the fallback of the default Spring DataSource in this case.
In the scenario where the host application has defined a DataSource but they misnamed it, there will be an exception as the Bean is retrieved by class in the else part and there will be more than one DataSource Beans (including the Spring default) so Spring will throw an exception.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
```diff
@@ -49,6 +54,18 @@ class ConfiguredApiRepository {
     log.debug("Query Execution time in ms: {}", stopwatch.time)
     return result
   }
+
+  private fun populateJdbcTemplate(dataSourceName: String): NamedParameterJdbcTemplate {
+    val dataSource = if (context.containsBean(dataSourceName)) {
+      context.getBean(dataSourceName, DataSource::class) as DataSource
+    } else {
+      log.warn("No DataSource Bean found withe name: {}", dataSourceName)
```
stalwartsolutions, line 62: Looks like a typo: "withe"

stalwartsolutions (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
gavriil-g-moj: Actually I have tested this scenario and what happens is that if there is already a custom DataSource Bean in the context, then the default Spring DataSource is not created. 
So there will be no exception thrown if someone has mistyped their DataSource Bean name but their custom Bean will be used instead of the default one.
As the code stands this will log a warning which probably shouldn't.

PR #79, opened 2024-01-29
Feature/dpr2 335 print option

rossdowthwaite:
Add `classification` and `printable` to API reponse.

Not sure if i have missed something or require extra steps. Please advise. thanks


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/ProductDefinition.kt
```diff
@@ -11,4 +11,6 @@ data class ProductDefinition(
   val dataset: List<Dataset> = emptyList(),
   val report: List<Report> = emptyList(),
   val policy: List<Policy> = emptyList(),
+  val classification: String? = null,
+  val printable: Boolean? = false,
```
stalwartsolutions, line 15: I believe these need to be added to the controller VariantDefinition, rather than here.

stalwartsolutions (approved):

stalwartsolutions (approved):

PR #81, opened 2024-02-02
Dpr2 341 Formula interpolation of dataset fields

gavriil-g-moj:
Implementation of formula interpolation for data fields.


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/FormulaEngine.kt
```diff
@@ -0,0 +1,30 @@
+package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.service
+
+import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.model.ReportField
+
+class FormulaEngine(private val reportFields: List<ReportField>) {
+
+  fun applyFormulas(row: Map<String, Any>): Map<String, Any> =
```
stalwartsolutions, line 7: I like the separation of concerns here. 👍 

Is there a possibility here of using a dataset's parameter (e.g. ${type_code}) which isn't normally in the returned results, as it doesn't have a counterpart field in the displayed report?

I feel like that might be a useful requirement. What do you think?

stalwartsolutions (approved):

PR #88, opened 2024-02-09
DPR2-323 make_url formula interpolation

gavriil-g-moj:
PR to allow for make_url formula interpolation.


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/ConfiguredApiService.kt
```diff
@@ -21,6 +22,7 @@ import java.time.format.DateTimeParseException
 class ConfiguredApiService(
   val productDefinitionRepository: ProductDefinitionRepository,
   val configuredApiRepository: ConfiguredApiRepository,
+  @Value("\${env:#{null}}") val env: String? = null,
```
stalwartsolutions, line 25: Does this property need to be documented so the consumers know it's there?
Also, is the name descriptive enough? In our Helm config for example, in the test values we'll have a value "ENV=dev" (to match the DPS env), which will look like a mistake.

gavriil-g-moj: Very good points, I meant to write the documentation thanks for the reminder.
Also regarding the env variable we decided to call it `url_env_suffix` which I had documented in the Jira and wrongly named `env` here. Thanks for bringing this up, I will fix it.

gavriil-g-moj: I have now made the changes.

stalwartsolutions (approved):

PR #100, opened 2024-02-28
DPR2-267 Fall back to dataset display if report field display is empty

gavriil-g-moj:
Fall back to dataset display if report field display is empty

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/SchemaField.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/SchemaField.kt
@@ -3,4 +3,5 @@ package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.model
 data class SchemaField(
   val name: String,
   val type: ParameterType,
+  val display: String,
 )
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/SchemaField.kt
```diff
@@ -3,4 +3,5 @@ package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.model
 data class SchemaField(
   val name: String,
   val type: ParameterType,
+  val display: String,
```
stalwartsolutions, line 6: Should this be optional? Or is it populated in the existing DPDs?

stalwartsolutions (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/SchemaField.kt
gavriil-g-moj: I checked the existing DPDs in the definitions repository and they have this field populated.

PR #105, opened 2024-03-04
[DPR2-423] Simplify getting latest published version

stalwartsolutions:


gavriil-g-moj (approved):
LGTM

.circleci/config.yml
```diff
@@ -18,8 +18,6 @@ jobs:
       name: hmpps/java
       tag: "19.0"
     steps:
-      - setup_remote_docker:
-          version: 20.10.14
       - checkout
```
gavriil-g-moj, line 21: Was this step not needed?

gavriil-g-moj: Or was this removed due to the deprecated image you mentioned causing the build to fail?

stalwartsolutions: The remote docker step was needed to support testcontainers, as it used docker for the PSQL image.
I realised that the deprecation was regarding remote docker, which explained why only the API lib build was failing.

.github/workflows/publish.yml
```diff
@@ -17,9 +17,6 @@ jobs:
   publish-package:
     runs-on: ubuntu-latest
 
-    permissions:
```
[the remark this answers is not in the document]
stalwartsolutions: The permissions step was copied from the UI library, which writes the published version back to the repo. I avoided this in the API lib, as it meant having less restrictions on main.

PR #123, opened 2024-03-21
[DPR2-494] Reconcile code to schema

stalwartsolutions:
Also removed remnants of testcontainers.


src/main/resources/data-product-definition-schema.json
```diff
@@ -1,298 +0,0 @@
-{
```
gavriil-g-moj, line 1: I think the library might be a better place for the schema instead of the definitions repository as it makes it visible to the other teams that use the library .
The definitions repository can link it in the Readme.
Although there might be an issue with validating the definitions in that way

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/WordWrap.kt
```diff
@@ -1,5 +1,14 @@
 package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.model
 
-enum class WordWrap {
+import com.google.gson.annotations.SerializedName
+
+enum class WordWrap() {
+  @SerializedName("none")
```
gavriil-g-moj, line 6: I have a feeling the @SerialiazedName annotation here has no effect. 
For example in the [this test](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/pull/123/files#diff-f9bb5eb2690fbc480103c00c739565d9eff660c0a4aaeac24ffb81a1cd20fdf4R270) the definition is taken from the `productDefinition.json` file where all instances of "wordwrap" which have a value are "None" with uppercase 'n'. 
Whereas the @SerializedName here defines "none" - lowercase 'n'. 
This is theory should have failed deserialisation but it works. 
This means this annotation has no effect and what determines the deserialisation is the actual enum value name.

stalwartsolutions: Interesting! I'll double check if all of these are needed. In particular the "break-words" one is not just a casing difference, so I'll add coverage to ensure this is acting as expected.

src/main/resources/data-product-definition-schema.json
stalwartsolutions: I don't think we're currently referring to the schema in the library anywhere in the documentation. So the fact that Aldo found it was purely him taking the initiative.

If we want to allow other users to see, I don't see any problem with adding a link to this repo from the library's docs.

In the library, this file isn't used, so it will go stale very fast. Here it will need to be kept up to date, otherwise the definitions will fail validation.

gavriil-g-moj (approved):
LGTM

gavriil-g-moj (changes requested):
Before merging could you please update the changelog to mention the support of the new fields.

gavriil-g-moj (approved):

Issue #111, opened 2024-03-07
Support for `boolean` fields


xoen:
Hello,

I've tried to add a `boolean` field but [it doesn't seem to be supported at the moment](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/blob/42261914fdde20d135bd2accb781436a3714f104/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/ParameterType.kt).

This seems to contradict the [list of supported types](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/blob/42261914fdde20d135bd2accb781436a3714f104/src/main/resources/data-product-definition-schema.json#L98) in the definition schema - unless I'm looking at an unrelated file of course.

At the moment I'm using a `string` type, but filtering doesn't seem to work (I'm still playing with it so there may be a workaround)

xoen:
Update: The workaround I'm using to make filtering work for boolean fields is to convert these columns to text, e.g.

```
      [...]
      "query": "SELECT [...], TEXT(na.is_closed) AS is_closed, [...]
```

gavriil-g-moj:
Boolean fields are now supported since version 3.7.0
There are the relevant changes:
https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/commit/be419a71abe4cb3650d862ecbc4f8b2aad10c09b

xoen:
Ah, interesting, I may have a play with this later, thanks @gavriil-g-moj ☺️

xoen:
Hi @gavriil-g-moj,

I've played with the boolean field for a few minutes and I can confirm the type itself is supported.

**However, filtering doesn't seem to work**. I tried a couple of things in the `filter` section but no dice. The problem seems to be with the generated SQL query adding `lower(is_closed)` on the boolean field which doesn't work (but again as usual, it may be something wrong with my `definitions.json`!)

**Changes**
I've updated the SQL query to remove the conversion to TEXT:

```
SELECT [...], na.is_closed AS is_closed,
[...]
```

Changed the is_closed field to be a `boolean`:
```
[...]
          {
            "name": "is_closed",
            "type": "boolean"
          },
[...]
```

and this is the filter that I had:

```
[...]
          {
            "name": "$ref:is_closed",
            "display": "Closed non-association?",
            "sortable": true,
            "filter": {
              "type": "Radio",
              "default": "false",
[...]

PR #126, opened 2024-03-27
DPR2-518 Fixed issue with filtering error on Boolean fields.

gavriil-g-moj:
This PR addresses https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/issues/111.
These changes fix the error when trying to filter by Boolean fields.
TestContainers were reintroduced to reproduce the failing scenario in a test as this problem was DB specific since it was occurring only on Redshift and not locally when the in memory H2 DB was used.
Therefore, using TestContainers with Postgres replicated the failure.
As TestContainers increase the build time significantly they were used only for the required test and not across all tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,8 @@ jobs:
       name: hmpps/java
       tag: "19.0"
     steps:
+      - setup_remote_docker:
+          version: 20.10.14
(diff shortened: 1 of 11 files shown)


dominic-messenger-moj (approved):
LGTM

PR #127, opened 2024-03-28
Reverse CHANGELOG.md versions order

xoen:
Generally newer versions should be on top of the file.

As time passes and the number of versions grows the information about older versions is less and less important/relevant.


gavriil-g-moj (approved):
LGTM

xoen:
Closing this PR, recreating to see if we can work around the permission issues.

PR #139, opened 2024-05-01
Dpr2 666 execute query api

gavriil-g-moj:
The following PR is creating the asynchronous execute statement API using Redshift data.
Note: This is pending cross account connectivity so the call will not work until this is resolved.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ConfiguredApiController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ConfiguredApiController.kt
@@ -113,6 +113,66 @@ class ConfiguredApiController(val configuredApiService: ConfiguredApiService) {
     }
   }
 
+  @GetMapping("/async/reports/{reportId}/{reportVariantId}")
+  @Operation(
+    description = "Executes asynchronously the dataset query for the given report and returns the execution ID. " +
+      "This is the asynchronous non-paginated version of the /reports/{reportId}/{reportVariantId} API.",
+    security = [ SecurityRequirement(name = "bearer-jwt") ],
+    responses = [
+      ApiResponse(
+        headers = [
+          Header(
+            name = NO_DATA_WARNING_HEADER_NAME,
+            description = "Provides additional information about why no data has been returned.",
+          ),
+        ],
+      ),
+    ],
+  )
+  fun asyncConfiguredApiExecuteQuery(
+    @RequestParam sortColumn: String?,
+    @RequestParam(defaultValue = "false") sortedAsc: Boolean,
+    @Parameter(
+      description = FILTERS_QUERY_DESCRIPTION,
+      example = FILTERS_QUERY_EXAMPLE,
+    )
+    @RequestParam
+    filters: Map<String, String>,
+    @PathVariable("reportId") reportId: String,
+    @PathVariable("reportVariantId") reportVariantId: String,
+    @Parameter(
+      description = ReportDefinitionController.DATA_PRODUCT_DEFINITIONS_PATH_DESCRIPTION,
+      example = ReportDefinitionController.DATA_PRODUCT_DEFINITIONS_PATH_EXAMPLE,
+    )
+    @RequestParam("dataProductDefinitionsPath", defaultValue = ReportDefinitionController.DATA_PRODUCT_DEFINITIONS_PATH_EXAMPLE)
+    dataProductDefinitionsPath: String? = null,
+    authentication: Authentication,
+  ): ResponseEntity<String> {
+    return try {
+      ResponseEntity
+        .status(HttpStatus.OK)
+        .body(
+          configuredApiService.validateAndExecuteStatementAsync(
+            reportId = reportId,
+            reportVariantId = reportVariantId,
+            filters = filtersOnly(filters),
+            sortColumn = sortColumn,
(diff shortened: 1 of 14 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ConfiguredApiController.kt
```diff
@@ -113,6 +113,66 @@ class ConfiguredApiController(val configuredApiService: ConfiguredApiService) {
...
+            reportId = reportId,
+            reportVariantId = reportVariantId,
+            filters = filtersOnly(filters),
+            sortColumn = sortColumn,
+            sortedAsc = sortedAsc,
+            userToken = if (authentication is DprAuthAwareAuthenticationToken) authentication else null,
+            dataProductDefinitionsPath = dataProductDefinitionsPath,
+          ),
+        )
+    } catch (exception: NoDataAvailableException) {
```
gavriil-g-moj, line 165: We might need to think if this is how we want to handle all exceptions on the frontend for this API or have some specific cases.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
```diff
@@ -7,11 +7,18 @@ import org.springframework.context.ApplicationContext
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate
 import org.springframework.stereotype.Service
+import software.amazon.awssdk.services.redshiftdata.RedshiftDataClient
+import software.amazon.awssdk.services.redshiftdata.model.ExecuteStatementRequest
+import software.amazon.awssdk.services.redshiftdata.model.ExecuteStatementResponse
+import software.amazon.awssdk.services.redshiftdata.model.SqlParameter
```
stalwartsolutions, line 13: Do you think it's worth separating this into two repositories? Possibly with a shared parent or helper class?

They're two quite different mechanisms, using two different sources.

src/test/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepositoryTest.kt
```diff
@@ -2,15 +2,22 @@ package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data
```
stalwartsolutions, line 2: Is it feasible to add an integration test for this too? To ensure the full path from the API to the repo is happy?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
gavriil-g-moj: Indeed I had thought of that and decided to put them in the same repository initially and confirm we can call this successfully first and then separate them.
But I will look into this again to make the separation from the beginning.

src/test/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepositoryTest.kt
gavriil-g-moj: I have tried this unsuccessfully as the only container I have found to use for testing is LocalStack.
I used [this with testcontainers](https://java.testcontainers.org/modules/localstack/) and there were two issues.
First of all it seemed unreliable as when I replaced Postgress with LocalStack the test we have in ConfiguredApiRepositoryTestContainersTest to catch a Redshift specific error would pass if I changed the code back to what was causing that bug. This test failed with the Postgres container as it should.
And secondly, which is the main blocker for that, running this test locally would not work as authentication fails when trying to connect to that LocalStack container.
I have been trying to test the Controller with a mocked service to confirm the controller calls the service but there is an issue with @MockBean not replacing the service with the mocked one in the Spring @WebMvcTest.
I will spend some more time on this and see if I can find an alternative.

stalwartsolutions: OK, thanks. If it's going to be a pig, then don't worry for this PR.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
gavriil-g-moj: The repositories have been separated and inheriting from a common helper class.

src/test/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepositoryTest.kt
gavriil-g-moj: I spent some more time but this test still has issues. 
We could merge this PR and look into this further as part of future changes.

stalwartsolutions (approved):

PR #182, opened 2024-06-10
Dpr2 817 Fix Redshift Error When Passing Filters

gavriil-g-moj:
These changes fix the issue due to which the Redshift async query was failing when the parameters' values were passed using the requestBuilder.parameters method from the Redshift Data Api.
Redshift Data API does not allow this inside CREATE EXTERNAL TABLE commands and throws this error:
`"ERROR: there is no parameter $1"`
The values are now replaced directly into the query string of the filter_ CTE.
Since this is isolated in the filter_ CTE it is not a concern for SQL injection.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
@@ -74,7 +74,7 @@ class ConfiguredApiRepository : RepositoryHelper() {
 
   private fun buildPreparedStatementNamedParams(filters: List<Filter>): MapSqlParameterSource {
     val preparedStatementNamedParams = MapSqlParameterSource()
-    filters.filterNot { it.type == FilterType.BOOLEAN }.forEach { preparedStatementNamedParams.addValue(it.getKey(), it.value.lowercase()) }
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ConfiguredApiRepository.kt
```diff
@@ -74,7 +74,7 @@ class ConfiguredApiRepository : RepositoryHelper() {
 
   private fun buildPreparedStatementNamedParams(filters: List<Filter>): MapSqlParameterSource {
     val preparedStatementNamedParams = MapSqlParameterSource()
-    filters.filterNot { it.type == FilterType.BOOLEAN }.forEach { preparedStatementNamedParams.addValue(it.getKey(), it.value.lowercase()) }
+    filters.filterNot { it.type == FilterType.DYNAMIC }.filterNot { it.type == FilterType.BOOLEAN }.forEach { preparedStatementNamedParams.addValue(it.getKey(), it.value.lowercase()) }
```
gavriil-g-moj, line 77: There is no point adding the dynamic filters as prepared statement parameters as their value is already replaced in the [filters_ CTE string ](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/blob/DPR2-817_redshift_error_when_passing_filters/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/RepositoryHelper.kt#L89). 
Also the query is using ILIKE so there is no reason to lowercase the value either.

tom-ogle-moj (approved):

PR #195, opened 2024-06-24
[DPR2-731] Allow dates (as well as datetimes) to be used as a source for format_date

stalwartsolutions:


koladeadewuyi-moj (approved):

gavriil-g-moj (approved):

gavriil-g-moj:
Thanks for fixing this. Interesting, this means there are dates of type Date in Redshift, other than Timestamp which was the only time we had until now.

PR #196, opened 2024-06-24
Dpr2 796 Filters with differt datasources

gavriil-g-moj:
Allow filters to use queries from different datasets other than the main report dataset.
The scope of this change is for the static options returned from the report definition endpoint.
Policies are ignored for these queries.


dominic-messenger-moj (approved):
All good

stalwartsolutions (approved):

PR #203, opened 2024-06-26
Dpr2 761 cancel statement

gavriil-g-moj:
Added the endpoint to cancel asynchronous statements which are still running.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ConfiguredApiController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ConfiguredApiController.kt
@@ -14,6 +14,7 @@ import org.springframework.http.HttpStatus
 import org.springframework.http.ResponseEntity
 import org.springframework.security.core.Authentication
 import org.springframework.validation.annotation.Validated
+import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.RequestParam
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ConfiguredApiController.kt
```diff
@@ -234,6 +236,50 @@ class ConfiguredApiController(val configuredApiService: ConfiguredApiService) {
...
+  @DeleteMapping("/reports/{reportId}/{reportVariantId}/statements/{statementId}")
+  @Operation(
+    description = "Cancels the execution of a running query.",
+    security = [ SecurityRequirement(name = "bearer-jwt") ],
+    responses = [
+      ApiResponse(
+        headers = [
+          Header(
+            name = NO_DATA_WARNING_HEADER_NAME,
+            description = "Provides additional information about why no data has been returned.",
```
stalwartsolutions, line 248: Is this header still relevant?

stalwartsolutions (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ConfiguredApiController.kt
gavriil-g-moj: Good point, it seems not. 
The NoDataAvailableException is thrown only when the caseloads are requested which in turn takes place in the policy engine.
For cancelling a query this is not relevant, as well as for other cases like getting the status, getting the already stored results of async queries and maybe more.
I will remove this.

stalwartsolutions (approved):

stalwartsolutions (approved):

PR #206, opened 2024-07-01
[DPR2-779] Add section property for list-section report templates

stalwartsolutions:


gavriil-g-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/Specification.kt
```diff
@@ -3,4 +3,5 @@ package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.model
 data class Specification(
   val template: String,
   val field: List<ReportField>,
+  val section: List<String> = emptyList(),
```
gavriil-g-moj, line 6: I think this will not work to default to empty list when deserialising from JSON and the mapper (`specification.section.map...`) will throw a Null Pointer Exception in the case when there is no section in the DPD.

PR #234, opened 2024-07-25
[DPR2-778] Lazy loading of summary data

stalwartsolutions:
- Added lazy loading of summary data if not yet created (works for both RedShift and Athena, but should only be used for Athena).
- Added use of coroutines to allow app to wait for table creation without blocking.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/ConfiguredApiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/ConfiguredApiService.kt
@@ -35,6 +35,7 @@ class ConfiguredApiService(
   val configuredApiRepository: ConfiguredApiRepository,
   val redshiftDataApiRepository: RedshiftDataApiRepository,
   val athenaApiRepository: AthenaApiRepository,
+  val tableIdGenerator: TableIdGenerator,
   val datasetHelper: DatasetHelper,
   @Value("\${URL_ENV_SUFFIX:#{null}}") val env: String? = null,
 ) {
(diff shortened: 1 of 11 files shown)


tom-ogle-moj (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/ConfiguredApiService.kt
```diff
@@ -164,38 +168,43 @@ class ConfiguredApiService(
...
+    val tableSummaryId = tableIdGenerator.getTableSummaryId(tableId, summaryId)
+
+    // Request data from the summary table.
+    // If it doesn't exist, create it (waiting for creation to complete).
+    val results = try {
+      redshiftDataApiRepository.getFullExternalTableResult(tableSummaryId)
+    } catch (e: BadSqlGrammarException) {
+      if (e.message?.contains("table or view does not exist") == true) {
+        requestRepo.createSummaryTable(productDefinition.datasource, tableId, summaryId, dataset)
+        redshiftDataApiRepository.getFullExternalTableResult(tableSummaryId)
```
gavriil-g-moj, line 194: I have a concern that this will most likely be returning no data as this line will be executed immediately before the above line for the table creation finishes.

stalwartsolutions: I believe the whole thread is suspended while it waits for the response, since it's all suspendable. But I can add a delay or such to the call to test this out?

gavriil-g-moj (approved):
Approving after our discussion with Stu and this will be confirmed in Dev testing.

PR #252, opened 2024-08-02
Dpr2 945 athena workgroup

gavriil-g-moj:
Change the Athena query configuration to make use of the following workgroup which has been setup:
`dpr-generic-athena-workgroup`
The results of the Athena queries will be stored in:
[s3://dpr-working-development/dpr](https://eu-west-2.console.aws.amazon.com/s3/buckets/dpr-working-development/dpr/)
Providing client configuration in the library for the S3 location will have no effect since the workgroup has been setup to override client configuration.
See below:
https://github.com/ministryofjustice/modernisation-platform-environments/blob/main/terraform/environments/digital-prison-reporting/modules/athena_workgroups/variables.tf#L24

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/AthenaApiRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/AthenaApiRepository.kt
@@ -7,7 +7,6 @@ import software.amazon.awssdk.services.athena.model.AthenaError
 import software.amazon.awssdk.services.athena.model.GetQueryExecutionRequest
 import software.amazon.awssdk.services.athena.model.QueryExecutionContext
 import software.amazon.awssdk.services.athena.model.QueryExecutionStatus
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/AthenaApiRepository.kt
```diff
@@ -73,13 +72,10 @@ class AthenaApiRepository(
       .database(datasource.database)
       .catalog(datasource.catalog)
       .build()
-    val resultConfiguration = ResultConfiguration.builder()
-      .outputLocation("s3://$s3location/$tableId/")
```
gavriil-g-moj, line 77: Providing client configuration in the library for the S3 location will have no effect since the workgroup has been setup to override client configuration.
See below:
https://github.com/ministryofjustice/modernisation-platform-environments/blob/main/terraform/environments/digital-prison-reporting/modules/athena_workgroups/variables.tf#L24

tom-ogle-moj (approved):

PR #280, opened 2024-08-16
[DPR2-000] Fix security issue

stalwartsolutions:
The overnight security check was flagging an issue with one of the spring dependencies.
To fix this, I upgraded the HMPPS Kotlin plugin, but it failed due to a dependency on Java 21.
To fix this, I upgraded the Java version, but it failed due to not being supported by the Gradle plugin.
To fix this, I upgraded the Gradle plugin version in IDEA and the wrapper, but this would fail in CircleCI due to the image version.
To fix this, I  upgraded the image to the correct Java version in the CircleCI builds.

I also implemented use of the HMPPS starter dependency (like the current template version). No issues.


gavriil-g-moj (dismissed):

gavriil-g-moj:
Nice work upgrading Java and fixing this security warning.
Maybe add these changes to the Changelog as a new major release?

gavriil-g-moj (approved):

PR #343, opened 2024-09-20
DPR2-1129 added new metrics data API

gavriil-g-moj:
The following changes create the new metrics data API.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/ReportDefinitionService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/ReportDefinitionService.kt
@@ -21,6 +21,7 @@ class ReportDefinitionService(
   ): List<ReportDefinitionSummary> {
     return productDefinitionRepository.getProductDefinitions(dataProductDefinitionsPath)
       .map { summaryMapper.map(it, renderMethod) }
+      // We might want to remove this filtering if we want to allow an empty list of variants for DPDs with metrics only
       .filter { it.variants.isNotEmpty() }
   }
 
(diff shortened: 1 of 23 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/ReportDefinitionService.kt
```diff
@@ -21,6 +21,7 @@ class ReportDefinitionService(
   ): List<ReportDefinitionSummary> {
     return productDefinitionRepository.getProductDefinitions(dataProductDefinitionsPath)
       .map { summaryMapper.map(it, renderMethod) }
+      // We might want to remove this filtering if we want to allow an empty list of variants for DPDs with metrics only
       .filter { it.variants.isNotEmpty() }
```
gavriil-g-moj, line 25: This will not allow using empty report variants for DPDs which contain only metrics.
We might want to remove this filter.

stalwartsolutions: This is to avoid returning reports with no variants when the render method of none of the variants matche. Perhaps an "or it.dashboards.isNotEmpty" would do the job?

gavriil-g-moj: Yes that's a good idea.

gavriil-g-moj: I have made this change.

stalwartsolutions (approved):

PR #413, opened 2024-10-30
[DPR2-1355] Definitions API Auth

stalwartstu:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ClientDataProductDefinitionsRepository.kt
```diff
@@ -21,9 +27,14 @@ class ClientDataProductDefinitionsRepository(
       path?.let { path -> cache.getIfPresent(path) }
     }
     cachedDefinitions?.let { return it }
+
+    val authentication = SecurityContextHolder.getContext().authentication as DprAuthAwareAuthenticationToken
```
gavriil-g-moj, line 31: If the token is required there is also the option to pass it through to the getProductDefinitions function all the way from the Controller.

gavriil-g-moj: This would make testing this method easier if there are no static dependencies.
But it seems the tests are passing?

gavriil-g-moj: In terms of when this method is stubbed in integration tests it doesn't cause a failure since we are using Wiremock and as long as the request method and path match, it intercepts the request.

gavriil-g-moj: There is a withHeader method but since we haven't specified it, Wiremock ignores the headers.

gavriil-g-moj: As this doesn't seem to be causing an issue at the moment I am happy to approve the PR and we can make these changes in the future if/when needed. 
Up to you.

gavriil-g-moj (approved):
Approving as this doesn't seem to be causing an issue at the moment. We can make these changes in the future if/when needed or if you choose to make them now I can re-review and reapprove.

PR #444, opened 2024-11-18
[DPR2-1363] Dashboard/interactive filters

stalwartstu:
- Added filters to dashboards.
- Extracted dashboard definition mapping.
- Added dashboard summary for getDefinitions endpoint (replacing full dashboard definition).
- Added interactive filters for reports.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/DataApiAsyncController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/DataApiAsyncController.kt
@@ -323,12 +323,12 @@ class DataApiAsyncController(val asyncDataApiService: AsyncDataApiService, val f
       .status(HttpStatus.OK)
       .body(
         asyncDataApiService.getStatementResult(
-          tableId,
-          reportId,
-          reportVariantId,
-          dataProductDefinitionsPath,
-          selectedPage,
-          pageSize,
+          tableId = tableId,
+          reportId = reportId,
+          reportVariantId = reportVariantId,
+          dataProductDefinitionsPath = dataProductDefinitionsPath,
+          selectedPage = selectedPage,
+          pageSize = pageSize,
           filters = filterHelper.filtersOnly(filters),
         ),
       )
@@ -364,13 +364,13 @@ class DataApiAsyncController(val asyncDataApiService: AsyncDataApiService, val f
       .status(HttpStatus.OK)
       .body(
         asyncDataApiService.getDashboardStatementResult(
-          tableId,
-          reportId,
-          dashboardId,
-          dataProductDefinitionsPath,
-          selectedPage,
-          pageSize,
-          filters,
+          tableId = tableId,
+          reportId = reportId,
+          dashboardId = dashboardId,
+          dataProductDefinitionsPath = dataProductDefinitionsPath,
+          selectedPage = selectedPage,
+          pageSize = pageSize,
+          filters = filterHelper.filtersOnly(filters),
         ),
       )
   }
@@ -399,11 +399,11 @@ class DataApiAsyncController(val asyncDataApiService: AsyncDataApiService, val f
       .status(HttpStatus.OK)
       .body(
         asyncDataApiService.getSummaryResult(
-          tableId,
-          summaryId,
-          reportId,
-          reportVariantId,
-          dataProductDefinitionsPath,
+          tableId = tableId,
+          summaryId = summaryId,
+          reportId = reportId,
+          reportVariantId = reportVariantId,
+          dataProductDefinitionsPath = dataProductDefinitionsPath,
           filters = filterHelper.filtersOnly(filters),
         ),
       )
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/DataApiAsyncController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/DataApiAsyncController.kt
@@ -121,6 +121,12 @@ class DataApiAsyncController(val asyncDataApiService: AsyncDataApiService, val f
     )
     @RequestParam("dataProductDefinitionsPath", defaultValue = ReportDefinitionController.DATA_PRODUCT_DEFINITIONS_PATH_EXAMPLE)
     dataProductDefinitionsPath: String? = null,
+    @Parameter(
+      description = FILTERS_QUERY_DESCRIPTION,
+      example = FILTERS_QUERY_EXAMPLE,
+    )
+    @RequestParam
+    filters: Map<String, String>,
     authentication: Authentication,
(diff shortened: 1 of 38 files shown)


rossdowthwaite (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/DataApiAsyncController.kt
```diff
@@ -121,6 +121,12 @@ class DataApiAsyncController(val asyncDataApiService: AsyncDataApiService, val f
     )
     @RequestParam("dataProductDefinitionsPath", defaultValue = ReportDefinitionController.DATA_PRODUCT_DEFINITIONS_PATH_EXAMPLE)
     dataProductDefinitionsPath: String? = null,
+    @Parameter(
+      description = FILTERS_QUERY_DESCRIPTION,
+      example = FILTERS_QUERY_EXAMPLE,
+    )
+    @RequestParam
+    filters: Map<String, String>,
```
gavriil-g-moj, line 129: Are the filters also mean to be applied at the dashboard execution or only at the results level?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/CommonDataApiService.kt
```diff
@@ -59,6 +60,13 @@ abstract class CommonDataApiService {
       ?.let { throw ValidationException("${AsyncDataApiService.MISSING_MANDATORY_FILTER_MESSAGE} ${it.display}") }
   }
 
+  protected fun checkMandatoryFiltersAreProvided(definition: SingleDashboardProductDefinition, filters: Map<String, String>, interactive: Boolean) {
+    definition.dashboardDataset.schema.field
+      .filter { it.filter?.interactive == interactive && it.filter.mandatory }
```
gavriil-g-moj, line 65: Won't this method throw an exception when called as part of the async dashboard execution when there are non interactive filters?

gavriil-g-moj: Or will the filters always be interactive for the async Dashboard execution?

gavriil-g-moj:
Also will addition the dashboard filter to the DashBoard and DashBoardDefinition classes be part of next changes?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/DataApiAsyncController.kt
stalwartstu: As per our Slack discussion - filters can be applied at either.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/CommonDataApiService.kt
stalwartstu: It will throw an exception if the filters are mandatory and not provided at the expected stage. For example, if an interactive filter is provided in the initial execution request, or if a non-interactive filter is provided in the results request.

stalwartstu: No it doesn't. I'm talking rubbish. 😓  It's just making mandatory filters are there in teh right context. One moment.

stalwartstu: Fixed!

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/DashboardDefinitionMapper.kt
```diff
@@ -0,0 +1,99 @@
...
+    ?.let { dynamicFilterOption ->
+      dynamicFilterOption.dataset
+        ?.let { dynamicFilterDatasetId ->
+          populateStaticOptionsForFilterWithDataset(
+            dynamicFilterOption,
+            allDatasets,
+            dynamicFilterDatasetId,
+            dynamicFilterOption.maximumOptions,
+          )
+        }
```
gavriil-g-moj, line 97: Are we assuming the dynamic filter options will be populated only by queries defined in the dynamic filter dataset and there will be no dynamic filters using the dashboard main dataset query?
It's fine if we do, just wanted to double check this was the intention.
For example, the report dynamic filters fall back to using the main report dataset query if there is no dataset referenced from the filter itself.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/CommonDataApiService.kt
```diff
@@ -52,16 +54,61 @@ abstract class CommonDataApiService {
...
-  protected fun validateAndMapFilters(definition: SingleReportProductDefinition, filters: Map<String, String>, reportFieldId: Set<String>? = null): List<ConfiguredApiRepository.Filter> {
+  private fun checkMandatoryFiltersAreProvided(
+    fieldFilters: Map<String, FilterDefinition?>,
+    interactive: Boolean?,
+    filters: Map<String, String>,
+  ) {
+    fieldFilters.keys
+      .firstOrNull {
+        fieldFilters[it] != null &&
+          fieldFilters[it]?.mandatory == true &&
```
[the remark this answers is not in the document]
stalwartstu: Yes. In that case this would filter them out (leaving nothing in the stream) and not throw the exception, I believe.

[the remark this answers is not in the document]
gavriil-g-moj: For example could in the DPD be filters defined but none of them as mandatory?
In this case this method would throw an exception.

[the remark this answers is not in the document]
gavriil-g-moj: Actually one moment, I'm having a second look at this.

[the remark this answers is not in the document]
stalwartstu: I don't follow, I'm afraid. In that case, the `fieldFilters[it]?.mandatory == true` check would filter it out, right?
And then the final check the `let` would have no results from `firstOrNull` and skip throwing the exception.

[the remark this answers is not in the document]
gavriil-g-moj: Yes, you are right.

gavriil-g-moj (approved):
LGTM

PR #475, opened 2024-11-29
[DPR2-1519] check dpd policy and return auth flag

djcostin:
- first cut of plugging in auth at a per dpd level, using the policy of the dpd to determine if a user has access

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 Below you can find the changes included in each release.
 
+## 7.2.3
+Added authorised flag to Report Summary Definition to determine if a user has access to a report definition.
+
 ## 7.2.2
 Added sortColumn and sortAsc query parameter to the async results endpoint to support the interactive journey.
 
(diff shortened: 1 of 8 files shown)


gavriil-g-moj:
This looks good to me. This brings a question regarding the individual report definition endpoint. If there is a call to the individual report definition endpoint, will this also need to cater for not returning a report to an anauthorised user?

gavriil-g-moj:
Also we would need to add the changelog documentation to this PR.

djcostin:
> This looks good to me. This brings a question regarding the individual report definition endpoint. If there is a call to the individual report definition endpoint, will this also need to cater for not returning a report to an anauthorised user?

yes I've got a separate ticket for that, which I'm currently working on https://dsdmoj.atlassian.net/browse/DPR2-1521

djcostin:
> Also we would need to add the changelog documentation to this PR.

I can update this now. Or does it make sense to update the changelog when all the auth changes are done ?

gavriil-g-moj:
> Reference in

Maybe as part of this PR you could mention in the changelog that the changes to the report definition summary were made to add the authorised flag and as part of the other ticket another changelog entry to capture the individual report definition endpoint changes.

gavriil-g-moj (approved):

PR #478, opened 2024-12-03
DPR2-1521 add auth to asyncDataAPI

djcostin:
- Add authorisation check against report policy for async endpoints

wanted to do some testing before implementing remainder of endpoints, still TODO

- sync endpoints
- dashboard based endpoints


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/DataApiAsyncController.kt
```diff
@@ -179,15 +190,30 @@ class DataApiAsyncController(val asyncDataApiService: AsyncDataApiService, val f
...
+        .body(
+          asyncDataApiService.getStatementStatus(
+            statementId = statementId,
+            reportId = reportId,
+            reportVariantId = reportVariantId,
+            userToken = if (authentication is DprAuthAwareAuthenticationToken) authentication else null,
+            dataProductDefinitionsPath,
+          ),
+        )
+    } catch (authException: UserAuthorisationException) {
```
gavriil-g-moj, line 212: It might be better to add this exception handling to ToolsApiExceptionHandler which has @RestControllerAdvice.
In that way you can handle the exception and determine the response only in one place instead of having to add it to individual endpoints.

djcostin: > It might be better to add this exception handling to ToolsApiExceptionHandler which has @RestControllerAdvice. In that way you can handle the exception and determine the response only in one place instead of having to add it to individual endpoints with try catch blocks.

I've updated that, thanks !

gavriil-g-moj:
LGTM. 
As part of future changes this will also be implemented for the single report definition endpoint, correct?

gavriil-g-moj (approved):
LGTM.
As part of future changes this will also be implemented for the single report definition endpoint, correct?

djcostin:
> LGTM. As part of future changes this will also be implemented for the single report definition endpoint, correct?

so this endpoint has the authorised flag / true / false, which will then drive the disabling of the various actions /links for a particular report definition. There will also be an option for the user to filter out the reports that they do not have access to

gavriil-g-moj:
> > LGTM. As part of future changes this will also be implemented for the single report definition endpoint, correct?
> 
> so this endpoint has the authorised flag / true / false, which will then drive the disabling of the various actions /links for a particular report definition. There will also be an option for the user to filter out the reports that they do not have access to

Yes this is for the /definitions endpoint which contains the summary of the report definitions. But wouldn't also the single report definition endpoint `/definitions/{reportId}/{variantId} `need to have a check not to return a definition for a non authorised request?

PR #485, opened 2024-12-05
[DPR2-1514] Add DDB definitions repo

stalwartstu:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/DynamoDbProductDefinitionRepository.kt
```diff
@@ -0,0 +1,37 @@
...
+        indexName = properties.categoryIndexName
+        keyConditionExpression = "${properties.categoryFieldName} = :${properties.categoryFieldName}"
+        expressionAttributeValues = attrValues
+        attributesToGet = listOf(properties.definitionFieldName)
+      }
+    }
+  }
+
+  override suspend fun getProductDefinitions(path: String?): List<ProductDefinition> {
+    return dynamoDbClient.query(getQueryRequest(properties, path ?: defaultPath)).items
```
gavriil-g-moj, line 32: Even though this is a coroutine and will not execute on the servlet thread handling the request, since the context is not switched, the servlet thread will not be released and it will be waiting until the blocking dynamo call completes.
So I guess this defeats the purpose if the reason of using coroutines was to release the thread handling the reqeust.
The blocking operation which is the dynamo call will need to be wrapped in a withContext block for the servlet thread to be released. For example:
  ```
withContext(Dispatchers.IO) {
dynamoDbClient.query(getQueryRequest(properties, path ?: defaultPath)).items
}
```

djcostin (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/DynamoDbProductDefinitionRepository.kt
gavriil-g-moj: I see, I hadn't realised the dynamamoDB client forces the suspend functions. 
This totally makes sense, I agree it's fine to move this way and not create a separate thread pool for an indexed read DynamoDB call.

gavriil-g-moj: Something else we should consider also is now that we have the DynamoDB client, for the individual report definition request, it's better to call Dynamo to get the single definition rather than getting all the definitions and then filtering in Kotlin by id to get the one requested.

PR #490, opened 2024-12-06
[DPR2-1521] refactor authCheck to only execute access level policies

djcostin:
- Updated remaining endpoints with authorisation policy check
- switched authCheck to only check access level policy types, was failing row level policies
- fixed tests adding default access level policy when needed


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/SyncDataApiService.kt
```diff
@@ -109,19 +114,30 @@ class SyncDataApiService(
...
     )
+    checkAuth(productDefinition, userToken)
     val policyEngine = PolicyEngine(productDefinition.policy, userToken)
     return Count(
       configuredApiRepository.count(
         filters = validateAndMapFilters(productDefinition, filters, null),
         query = productDefinition.reportDataset.query,
         reportId = reportId,
-        policyEngineResult = policyEngine.execute(),
+        policyEngineResult = policyEngine.execute(PolicyType.ROW_LEVEL),
```
stalwartstu, line 124: Do you also need Access here, too?

djcostin: ah good spot, actually should just be execute(), looking at it now

djcostin: updated that

stalwartstu (approved):

gavriil-g-moj:
The single report definition endpoint doesn't seem to have the auth check as part of these changes. (edited)

PR #552, opened 2025-01-23
[DPR2-1646] Child reports and ID searching refactor

stalwartstu:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/CommonDataApiService.kt
```diff
@@ -205,18 +208,15 @@ abstract class CommonDataApiService {
   }
 
   fun findFilterDefinitionFromFieldsAndDataset(fields: List<ReportField>?, dataset: Dataset, filterName: String): FilterDefinition? {
-    val fieldFilter = fields?.firstOrNull { it.filter != null && filterName == it.name.removePrefix(AsyncDataApiService.SCHEMA_REF_PREFIX) }?.filter
+    val fieldFilter = identifiedHelper.findOrNull(fields, filterName)?.filter
```
gavriil-g-moj, line 211: Would this work as intended?
Previously it was comparing the filter name with the field filter.
I think now it would compare the filterName with the field name?

gavriil-g-moj: Now it is finding the field based on the filter name and then taking the filter of that field.

stalwartstu: Previously it checked whether the field had a filter, then checked the filter name matched the field name - "it" here is the field.

Now it just checks the names match, since if there was more than one field with a matching name, you'd be having a bad day anyway! 😆

stalwartstu: Filter name here is a bit of a misnomer, it's really the filter's field name.,

gavriil-g-moj (approved):
LGTM

PR #561, opened 2025-01-30
Changed dashboard json structure

gavriil-g-moj:
Made the following changes to the dashboard json structure in preparation for supporting dashboards with lists:
- Moved "unit" from chart to column.
- Introduced new optional "aggregate" attribute of type "sum" to column.
- Added columns to metric as a list of full json column objects.
- Replaced chart columns value with a list of strings referencing the columns in the metric.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 Below you can find the changes included in each release.
 
+# 7.5.1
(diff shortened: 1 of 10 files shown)


gavriil-g-moj:
This PR is now obsolete since we moved ahead with a different dashboard definition.

PR #566, opened 2025-02-10
IR-778 Add ability to get the list of caseloads a user has and use this for policy checks

Mjwillis:
This allows to filter by a list of caseloads a user has access to:

https://github.com/ministryofjustice/hmpps-digital-prison-reporting-data-product-definitions-schema/pull/33

```json
"policy": [
    {
      "id": "caseloads",
      "type": "row-level",
      "action": ["establishment_code in (${caseloads})"],
      "rule": [
        {
          "effect": "permit",
          "condition": [
            {
              "exists": ["${caseloads}"]
            }
          ]
        }
      ]
    }
  ],
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 Below you can find the changes included in each release.
 
+# 7.6.1
+Support policy checks a list of available caseloads a user is allowed to access.
+
 # 7.6.0
 Dashboard definition response converts dataset parameters to filters.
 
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 Below you can find the changes included in each release.
 
+# 7.6.1
```
gavriil-g-moj, line 3: After having checked the published library versions on Sonatype, 7.7.0 was published from this branch so this should be 7.7.0 instead of 7.6.1.



<br class="Apple-interchange-newline">

PR #567, opened 2025-02-10
IR-778 Update change log to match version

Mjwillis:


CHANGELOG.md
```diff
@@ -1,7 +1,32 @@
 Below you can find the changes included in each release.
 
-# 7.6.1
-Support policy checks a list of available caseloads a user is allowed to access.
+# 7.7.0
```
gavriil-g-moj, line 3: Since the latest version is 7.8.0, it might be better to have that instead. 
Even though both of these versions contain only these changes.

gavriil-g-moj: Or it could even be
# 7.7.0 - 7.8.0

gavriil-g-moj (approved):

PR #578, opened 2025-02-25
IR-818 Refactor authorisation handling to use kotlin library and secure endpoints without use of default role

Mjwillis:
Integrate the new Koltin library support for multiple resource configurations.
Remove the "default" role protection and made it explicit so that other clients of the library are not clobbered by the default role.


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprResourceServerConfiguration.kt
```diff
@@ -0,0 +1,53 @@
...
+fun removeRolePrefix(listOfRoles: List<String>) = listOfRoles.map { it.replace("ROLE_", "") }
+
+private fun ResourceServerConfigurationCustomizerDsl.authorizeHttpRequests(roles: List<String>) {
+  authorizeHttpRequests {
+    authorize("/report/**", hasAnyRole(*roles.toTypedArray()))
+    authorize("/reports/**", hasAnyRole(*roles.toTypedArray()))
+    authorize("/definitions/**", hasAnyRole(*roles.toTypedArray()))
+    authorize("/statements/**", hasAnyRole(*roles.toTypedArray()))
+    authorize("/async/**", hasAnyRole(*roles.toTypedArray()))
+  }
```
mikehalmamoj, line 52: I think this means you aren't getting any unauthorized endpoints or other Kotlin lib defaults. However you'll need to define another filter chain in your application anyway (the incidents one) as suggested by @petergphillips:
```
@Bean
  fun hmppsSecurityFilterChain(http: HttpSecurity, customizer: ResourceServerConfigurationCustomizer): SecurityFilterChain =
    HmppsResourceServerConfiguration().hmppsSecurityFilterChain(http, customizer)
```

So maybe this is fine.

Mjwillis:
Will wait on https://github.com/ministryofjustice/hmpps-kotlin-lib/pull/64

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprResourceServerConfiguration.kt
```diff
@@ -6,19 +6,33 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
...
 @Configuration("dprResourceServerConfiguration")
 @ConditionalOnProperty(name = ["dpr.lib.user.role", "spring.security.oauth2.resourceserver.jwt.jwk-set-uri"])
 @AutoConfigureBefore(WebMvcAutoConfiguration::class)
-class ResourceServerConfiguration(
+class DprResourceServerConfiguration(
   private val caseloadProvider: CaseloadProvider,
   @Value("\${dpr.lib.user.role}") private val authorisedRole: String,
 ) {
 
+  @Order(1)
```
gavriil-g-moj, line 23: Is the intention of this to apply first before any other SecurityFilterChain defined by the clients of the library?
If that's the case would it be better for this to have the Ordered.LOWEST_PRECEDENCE value?
Or do we need to allow for more flexibility in terms of ordering in which case 1 would be fine.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprResourceServerConfiguration.kt
```diff
@@ -6,19 +6,33 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
...
+    http: HttpSecurity,
+    dprResourceServerCustomizer: ResourceServerConfigurationCustomizer,
+  ): SecurityFilterChain = HmppsResourceServerConfiguration().hmppsSecurityFilterChain(http, dprResourceServerCustomizer)
+
   @Bean
   fun resourceServerCustomizer() = ResourceServerConfigurationCustomizer {
-    anyRequestRole { authorisedRole }
     oauth2 { tokenConverter = DefaultDprAuthAwareTokenConverter(caseloadProvider) }
+    securityMatcher { listOf("/report/**", "/reports/**", "/definitions/**", "/statements/**", "/async/**") }
+    anyRequestRole { defaultRole = removeRolePrefix(authorisedRole) }
```
gavriil-g-moj, line 34: Previously the role was added without the prefix being stripped. 
Was there and change in the Kotlin libary which now requires the prefix to be stripped?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprResourceServerConfiguration.kt
gavriil-g-moj, line 23: Is the intention of this to apply first before any other SecurityFilterChain defined by the clients of the DPR library?
If that's the case would it be better for this to have the `Ordered.HIGHEST_PRECEDENCE` value?
Or do we need to allow for more flexibility in terms of ordering in which case 1 would be fine.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprResourceServerConfiguration.kt
Mjwillis: Yes for the ROLE_ needs to be striped or not added in the first place, this will support either

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprResourceServerConfiguration.kt
Mjwillis: Yes it comes first - @mikehalmamoj what do you think about the Ordered.HIGHEST_PRECEDENCE idea?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprResourceServerConfiguration.kt
Mjwillis: Yes for default

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprResourceServerConfiguration.kt
mikehalmamoj: Yes that is the intention so HIGHEST_PRECEDENCE would work fine. However as you point out this gives us a bit more flexibility. It's hard to guess the future so I prefer the flexibility - I think?

gavriil-g-moj (approved):

PR #586, opened 2025-03-04
IR-818 Fix resource server

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprResourceServerConfiguration.kt
```diff
@@ -28,9 +28,9 @@ class DprResourceServerConfiguration(
   ): SecurityFilterChain = HmppsResourceServerConfiguration().hmppsSecurityFilterChain(http, dprResourceServerCustomizer)
 
   @Bean
-  fun resourceServerCustomizer() = ResourceServerConfigurationCustomizer {
+  fun dprResourceServerCustomizer() = ResourceServerConfigurationCustomizer {
```
mikehalmamoj, line 31: This really needs to match the parameter name up on line 27 - Spring resorts to bean names if there are multiple beans of the same type, so the bean name and parameter name should match.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprResourceServerConfiguration.kt
```diff
@@ -28,9 +28,9 @@ class DprResourceServerConfiguration(
...
     oauth2 { tokenConverter = DefaultDprAuthAwareTokenConverter(caseloadProvider) }
-    securityMatcher { listOf("/report/**", "/reports/**", "/definitions/**", "/statements/**", "/async/**") }
+    securityMatcher { paths = listOf("/report/**", "/reports/**", "/definitions/**", "/statements/**", "/async/**") }
```
mikehalmamoj, line 33: One of the downsides of using a DSL is that it's "just code" - so we were declaring a list but it was being thrown away and not actually being used for the security matcher. The `paths` property of the DSL sets those paths in the security matcher.

Mjwillis (approved):
LGTM

PR #600, opened 2025-03-13
IR-938 Enhance DPR Api library to handle system tokens with user in context

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/UserPermissionProvider.kt
```diff
@@ -0,0 +1,53 @@
...
+
+  fun getUsersRoles(username: String): List<String> = manageUsersWebClient.get()
+    .uri("/users/$username/roles")
+    .header("Content-Type", "application/json")
+    .retrieve()
+    .bodyToMono(ROLES)
+    .block()!!.map { it.roleCode }
+
+  private fun getPrisonUsersCaseload(username: String): CaseloadResponse = manageUsersWebClient.get()
+    .uri("/prisonusers/$username/caseloads")
```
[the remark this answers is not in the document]
Mjwillis: yes

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DefaultUserPermissionProvider.kt
```diff
@@ -0,0 +1,53 @@
...
+
+  override fun getUsersRoles(username: String): List<String> = manageUsersWebClient.get()
+    .uri("/users/$username/roles")
+    .header("Content-Type", "application/json")
+    .retrieve()
+    .bodyToMono(ROLES)
+    .block()!!.map { it.roleCode }
+
+  override fun getPrisonUsersCaseload(username: String): CaseloadResponse = manageUsersWebClient.get()
+    .uri("/prisonusers/$username/caseloads")
```
gavriil-g-moj, line 42: Since we are authenticating with system credentials this is a system user here.
Does this mean the `username` is a system username and not the one of a real user?
Therefore are we retrieving the role and caseloads of a system user?
In our meeting yesterday it was mentioned there is a way for the actual user to be inferred when use of system tokens is made. Is this the case here and if so how do this work?

Mjwillis: This is the real user here NOT the system user - the calling API will pass in the username as part of the token generation - this is standard practice.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/ClientDataProductDefinitionsRepository.kt
```diff
@@ -31,6 +31,7 @@ class ClientDataProductDefinitionsRepository(
 
     val respEntity: ResponseEntity<List<ProductDefinition>>? = path?.let {
       val headers: MultiValueMap<String, String> = HttpHeaders()
+      // TODO: should not be passing the existing token onwards - should create a new one
```
Mjwillis, line 34: This can be looked at later in a future change

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/policyengine/Condition.kt
```diff
@@ -35,8 +35,8 @@ data class Condition(
     authToken: DprAuthAwareAuthenticationToken?,
     matchList: List<String>,
   ): Boolean {
-    val userRoles = authToken?.authorities?.map { it.authority }
-    return userRoles?.any { it in matchList } ?: false
+    val userRoles = authToken?.getRoles()?.removeRolePrefix()
```
Mjwillis, line 38: Want to remove any issues with either having or not having a ROLE_ prefix, both should work

gavriil-g-moj (approved):
Looks good to me.

PR #603, opened 2025-03-19
IR-938 DprSystemAuthAwareAuthenticationToken to handle null userName gracefully

Mjwillis:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprSystemAuthAwareAuthenticationToken.kt
```diff
@@ -23,28 +23,34 @@ class DprSystemAuthAwareAuthenticationToken(
 
   override fun getActiveCaseLoadId(): String? {
     if (this.activeCaseload == null) {
-      this.activeCaseload = userPermissionProvider.getActiveCaseloadId(this.userName!!)
+      this.activeCaseload = this.userName?.let {
+        userPermissionProvider.getActiveCaseloadId(it)
+      } ?: "NO_CASELOAD"
```
gavriil-g-moj, line 28: This might be better not to assign the "NO_CASELOAD" string in case the username is null.
So activeCaseload would also be null in this case.
There are cases for example in the Policy Engine where rules with exists conditions can be defined:
`"rule": [
        {
          "effect": "permit",
          "condition": [
            {
              "exists": ["${caseload}"]
            }
          ]
        }
      ]`
In this case this exist condition would be true if the caseload was defaulted to "NO_CASELOAD".
However, this would not be correct in this case.

Mjwillis: @gavriil-g-moj yes agree - pushed change

gavriil-g-moj (approved):

PR #626, opened 2025-05-19
DPR2-1974 add alert category cache

djcostin:
add alert category / code cache to be used for to populate dropdown on the front end


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/alert/AlertCategoryRepository.kt
```diff
@@ -0,0 +1,99 @@
...
+class AlertCategoryRepository(
+  override val athenaClient: AthenaClient,
+  override val tableIdGenerator: TableIdGenerator,
+  @Value("\${dpr.lib.redshiftdataapi.athenaworkgroup:workgroupArn}")
+  override val athenaWorkgroup: String,
+  val athenaQueryHelper: AthenaQueryHelper = AthenaQueryHelper(),
+) : AthenaApiRepository(athenaClient, tableIdGenerator, athenaWorkgroup) {
+
+  companion object {
+    const val NOMIS_CATALOG = "nomis"
```
gavriil-g-moj, line 27: Since the NOMIS_CATALOG and DIGITAL_PRISON_REPORTING_DB are already defined constants in EstablishmentsToWingsRepository maybe they can be referenced here instead of having to be redefined?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/establishmentsAndWings/AthenaQueryHelper.kt
```diff
@@ -0,0 +1,36 @@
+package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.establishmentsAndWings
+
+import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.QUERY_ABORTED
+import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.QUERY_FAILED
+import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.QUERY_FINISHED
+import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.model.redshiftdata.StatementExecutionStatus
+
+class AthenaQueryHelper {
```
gavriil-g-moj, line 8: Since `AthenaApiRepository` is an existing parent class of both `AlertCategoryRepository` and `EstablishmentsToWingsRepository` can the `waitForQueryToComplete` method be moved there instead?
This will make a new helper class redundant.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/alert/AlertCategoryCacheService.kt
```diff
@@ -0,0 +1,41 @@
...
+
+  fun getAlertCodesCacheIfNeeded(): Map<String, List<AlertCategory>> {
+    if (alertCodesCache.size() == 0L) {
+      log.debug("AlertCategory cache was empty.")
+      refresh()
+    }
+    return alertCodesCache.asMap()
+  }
+
+  fun refresh() {
```
gavriil-g-moj, line 28: Are the alert codes static data which will be retrieved only once or do they change?
If they change a periodic cache refresh would be needed.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/DefinitionMapper.kt
```diff
@@ -170,6 +174,10 @@ abstract class DefinitionMapper(
       ?: emptyList()
   }
 
+  private fun mapAlertsToFilterOptions(): List<FilterOption> = alertCategoryCacheService.getAlertCodesCacheIfNeeded()
+    .flatMap { alerts -> alerts.value.map { FilterOption(it.code, it.description) } }.toList()
```
gavriil-g-moj, line 178: Since only the code and description are returned to the FE is there a need to store the [domainCode](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/pull/626/files#diff-a11a176d7ab69d28eeff5bb62933c5fc1d3098dc7ccfb9669e9a584d0d14d5a9R4) in the AlertCategory?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/alert/AlertCategoryRepository.kt
djcostin: yeah I was looking at refactoring these into a common place, will look at that

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/establishmentsAndWings/AthenaQueryHelper.kt
djcostin: Yes that was my first initial thought but opted for composition over inheritance as its only used by these 2 classes and the existing parent class was already quite big but I can move it for simplicity

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/DefinitionMapper.kt
djcostin: I need to check on this as this was in the original query in the jira

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/alert/AlertCategoryCacheService.kt
djcostin: I'm not sure, its only used by 1 report as well that I'm aware of, but could look to add this later if need be

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/alert/AlertCategoryRepository.kt
gavriil-g-moj: This refactor can be done incrementally maybe so that we don't delay your changes?
If you prefer you could leave this for a next PR so that you can merge this one and test it first.

djcostin: thanks, I'm happy to do that

gavriil-g-moj (dismissed):
Agreed with @djcostin comments will be addressed in a next PR.

gavriil-g-moj (approved):
Agreed with @djcostin comments will be addressed in a next PR.

PR #645, opened 2025-06-24
DPR2-1860: sortDirection in DPDs

ReedSoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/CommonDataApiService.kt
```diff
@@ -34,13 +35,31 @@ abstract class CommonDataApiService(val identifiedHelper: IdentifiedHelper) {
...
+    val sortColumn = findSortColumn(sortColumn, productDefinition.reportDataset)
+    if (sortColumn == null) {
+      val (sortColumn, sortDirection) = calculateDefaultSortColumn(productDefinition)
+      val computedSortedAsc = if (sortedAsc != null && sortColumn != null) sortedAsc else sortDirection
+      return Pair(sortColumn, computedSortedAsc)
+    }
+    return Pair(
+      sortColumn,
+      sortedAsc
+        ?: (productDefinition.report.specification?.field?.firstOrNull { it.name == sortColumn }?.sortDirection === SortDirection.ASC),
```
gavriil-g-moj, line 60: I think this is not correct:
`it.name == sortColumn`
The specification fields contain the `$ref:` prefix in their name field. 
So this needs to be removed first for the comparison with the sortColumn query param.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/model/SortDirection.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.controller.model
+
+import com.fasterxml.jackson.annotation.JsonValue
+
+enum class SortDirection(@JsonValue val value: String) {
+  ASC("ASC"), // deserializer maybe - check lowercase
```
gavriil-g-moj, line 6: I think it's best to have the string value which needs to be specified in the DPD as lowercase.
This is consistent with how the rest of the enums are defined in the schema:
`https://github.com/ministryofjustice/hmpps-digital-prison-reporting-data-product-definitions-schema/blob/main/schema/data-product-definition-schema.json#L177C24-L177C28`

gavriil-g-moj: Either all lowercase or we could make the DPD accept any casing.
I would suggest all lowercase as it is simpler while accepting any casing is more like a nice to have.

src/test/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/integration/DataApiIntegrationTest.kt
```diff
@@ -62,6 +67,115 @@ class DataApiIntegrationTest : IntegrationTestBase() {
           }
   """.trimIndent()
 
+  /**
+   * Nested test class so we can add a couple entries to the mock db just for these tests
```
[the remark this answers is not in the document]
ReedSoftware: Thanks, this was because of the bug you spotted with the prefix.

src/test/resources/productDefinition.json
```diff
@@ -179,7 +179,8 @@
           "default" : "today(-1,months) - today()"
         },
         "sortable" : true,
-        "defaultsort" : true
+        "defaultsort" : true,
+        "sortDirection" : "ASC"
```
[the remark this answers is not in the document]
ReedSoftware: > Can we have this defined as lowercase here please, as this is the main casing we would need to support. This is consistent with how the other enum values are defined in the DPD josn.

Done

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/model/SortDirection.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.controller.model
+
+import com.fasterxml.jackson.annotation.JsonValue
+
+enum class SortDirection(@JsonValue val value: String) {
```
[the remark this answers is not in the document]
gavriil-g-moj: For example see the ParameterType data model:
https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/ParameterType.kt

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/model/SortDirection.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.controller.model
```
gavriil-g-moj, line 1: As mentioned [here](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/pull/645#discussion_r2177022110) this class should move to the `data.model` package.
And a separate class called SortDirectionDefinition as an example name can be created in the here in the `controller.model` package.
This is because we have been following a pattern on keeping the data model (model read from the DPD) separate from the controller model (definition response returned to the consumers of the API).
The data model have the SerializedName annotation e.g. @SerializedName("asc") like here to read the JSON value.
The controller model has the @JsonValue annotation in the constructor of the class to serialise the value as defined there.
See the [SummaryTemplate](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/SummaryTemplate.kt#L6) as an example of a data model and the corresponding [SummaryTemplate](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/model/SummaryTemplate.kt) as its corresponding controller model.

gavriil-g-moj (dismissed):
LGTM

gavriil-g-moj (approved):

PR #652, opened 2025-07-08
[DPR2-2007] feat: bundle in missing reports to definitions returned

ReedSoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/ReportDefinitionController.kt
```diff
@@ -43,7 +43,7 @@ class ReportDefinitionController(val reportDefinitionService: ReportDefinitionSe
       description = DATA_PRODUCT_DEFINITIONS_PATH_DESCRIPTION,
       example = DATA_PRODUCT_DEFINITIONS_PATH_EXAMPLE,
     )
-    @RequestParam("dataProductDefinitionsPath", defaultValue = DATA_PRODUCT_DEFINITIONS_PATH_EXAMPLE)
+    @RequestParam("dataProductDefinitionsPath")
```
gavriil-g-moj, line 46: I think this can be left here as is.
The missing reports would need to be checked even if the path is present.

gavriil-g-moj:
Also a suggestion regarding the commit messages. It's better when the message starts with the Jira number like `DPR2-2007` as it's much easier to identify the Jira when looking at the changes with Git Blame.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/DynamoDbProductDefinitionRepository.kt
```diff
@@ -16,47 +17,54 @@ class DynamoDbProductDefinitionRepository(
   identifiedHelper: IdentifiedHelper,
 ) : AbstractProductDefinitionRepository(identifiedHelper) {
   companion object {
-    const val DEFAULT_PATH = "definitions/prisons/orphanage"
+    val DEFAULT_PATHS = listOf(
```
gavriil-g-moj, line 20: This constant is not used.

gavriil-g-moj (dismissed):
Looks good to me. Only minor comment is that you can assert the definition path value too in the DynamoDbProductDefinitionRepositoryTest.

PR #889, opened 2026-04-23
Pdhd 140 non nomis users

ReedSoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DefaultUserPermissionProvider.kt
```diff
@@ -3,32 +3,44 @@ package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.security
...
 
-    return caseloadResponse.activeCaseload.id
+    return caseloadResponse.caseloads.sortedBy { it.id }.map { Caseload(it.id, it.name) }
   }
 
-  override fun getCaseloads(username: String): List<Caseload> {
-    val caseloadResponse = getPrisonUsersCaseload(username)
+  override fun getUserInfo(username: String): AuthUser {
+    val user = manageUsersWebClient.get()
+      .uri("/users/$username/me")
```
gavriil-g-moj: Are you sure this endpoint exists?
Looking at the manage-users Swagger docs I couldn't find this:
https://manage-users-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/user-controller/myDetails

gavriil-g-moj: Did you mean to call `users/${username}` endpoint instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/authentication/AuthUser.kt
```diff
@@ -0,0 +1,14 @@
...
+import uk.gov.justice.hmpps.kotlin.auth.AuthSource
+
+data class AuthUser(
+  val username: String,
+  val active: Boolean,
+  val name: String,
+  val authSource: AuthSource,
+  val userId: String,
+  val uuid: String,
+  val staffId: Int?,
```
gavriil-g-moj: This in the [Swagger docs](https://manage-users-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/user-controller/findUser) staffId is marked as deprecated and says `Deprecated, use userId instead`

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/authentication/AuthUser.kt
```diff
@@ -0,0 +1,14 @@
...
+  val activeCaseLoadId: String?,
```
gavriil-g-moj: In the [Swagger docs](https://manage-users-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/user-controller/findUser) activeCaseLoadId is marked as deprecated and says: `Deprecated, retrieve from prison API rather than manage users`

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprSystemAuthAwareAuthenticationToken.kt
```diff
@@ -24,7 +24,7 @@ class DprSystemAuthAwareAuthenticationToken(
   override fun getActiveCaseLoadId(): String? {
     if (this.activeCaseload == null) {
       this.activeCaseload = this.userName?.let {
-        userPermissionProvider.getActiveCaseloadId(it)
+        userPermissionProvider.getUserInfo(it).activeCaseLoadId
```
gavriil-g-moj: I don't think we should be getting the active caseload from the users/{username} endpoint. 
This has been marked as deprecated and suggests to use the prison API instead.
So we should be getting it from [/prisonusers/{username}/caseloads](https://manage-users-api-dev.hmpps.service.justice.gov.uk/swagger-ui/index.html#/user-caseloads-controller/getUserCaseloads)

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DefaultUserPermissionProvider.kt
djcostin: we should note use the /users/$username/me endpoint but the general /users/{username} one as we are passing a system token to it rather then a user token

src/test/resources/application-system-test.yml
```diff
@@ -36,8 +41,12 @@ management.endpoint:
 
 dpr:
   lib:
+    caseloads:
+      host: http://localhost:9999
```
djcostin: do we need this as its actually the same api as manage users, actually looks like it was already there

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DefaultUserPermissionProvider.kt
ReedSoftware: Yes I did, thank you.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/authentication/AuthUser.kt
ReedSoftware: I'll remove it then, we arent using it anyway

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprSystemAuthAwareAuthenticationToken.kt
ReedSoftware: Ok fair enough, will change this up

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DefaultUserPermissionProvider.kt
ReedSoftware: Yeah gavriil also spotted this, will update cheers

src/test/resources/application-system-test.yml
ReedSoftware: Possibly not, will look

gavriil-g-moj (approved):
LGTM.
We agreed to test the manage-users `/prisonusers/$username/caseloads` API response for external users with a system token to confirm if it is a 404.

ReedSoftware:
Confirmed its 404 for most normal usernames being queried. Possibly 400 if external users have symbols in, but something to followup with later as we're not sure of what format external users will take.

PR #911, opened 2026-05-01
Request scoped user info

ReedSoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/productCollection/ProductCollectionService.kt
```diff
@@ -2,15 +2,16 @@ package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.productCollection
...
 @Service
 class ProductCollectionService(
   val productCollectionRepository: ProductCollectionRepository,
+  val userRequestContext: UserRequestContext,
 ) {
-  fun getProductCollections(authToken: DprAuthAwareAuthenticationToken?): Collection<ProductCollectionSummary> {
+  fun getProductCollections(): Collection<ProductCollectionSummary> {
     // Set this to "NULL" if theres no values, which will fail the check, if a collection has any caseloadId attributes linked to it
-    val caseloadIds = authToken?.getCaseLoadIds()?.takeIf { it.isNotEmpty() } ?: listOf("NULL")
+    val caseloadIds = userRequestContext.getCaseLoadIds().takeIf { it.isNotEmpty() } ?: listOf("NULL")
```
gavriil-g-moj: I think `getCaseLoadIds().takeIf` is missing the `?`

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/UserRequestContext.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.web.context.WebApplicationContext
+import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.security.authentication.AuthUser
+
+@Component
+@Scope(value = WebApplicationContext.SCOPE_REQUEST, proxyMode = ScopedProxyMode.TARGET_CLASS)
+class UserRequestContext(
+  private val manageUsersClient: ManageUsersClient,
+) {
+  val authentication: DprSystemAuthAwareAuthenticationToken by lazy {
+    SecurityContextHolder.getContext().authentication as DprSystemAuthAwareAuthenticationToken
```
gavriil-g-moj: Will the token always be a DprSystemAuthAwareAuthenticationToken?
It might but it's probably better to be defensive here and throw the exception ourselves rather than let it break in an unexpected way, like NPE or a class cast exception.
Something like:
`  SecurityContextHolder.getContext().authentication as?DprSystemAuthAwareAuthenticationToken ?: throw IllegalStateException("Authentication was not of DprSystemAuthAwareAuthenticationToken type")`

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/UserRequestContext.kt
```diff
@@ -0,0 +1,25 @@
...
+  }
+  val prisonCaseloadData: CaseloadResponse by lazy { manageUsersClient.getCaseloads(authentication.userName!!) }
```
gavriil-g-moj: But some users like external users will not have caseloads. This will return a 400 from the manage-users API because the username is of email format.
And for other users it will return a 404.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/UserRequestContext.kt
```diff
@@ -0,0 +1,25 @@
...
+  val userRoles: List<String> by lazy { authentication.userName?.let { manageUsersClient.getUsersRoles(it) }.orEmpty() }
+  val userInfo: AuthUser by lazy { manageUsersClient.getUserInfo(authentication.userName!!) }
```
gavriil-g-moj: I think it's better to be consistent.
On one had for the userRoles we allow a null username and return an empty list.
On the other hand here and in the caseloads calls if the username is null it will throw NPE.
At least as a first step let's maybe use the elvis operator to throw our own exception with a custom message.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/establishmentsAndWings/EstablishmentsToWingsRepository.kt
```diff
@@ -21,7 +22,8 @@ class EstablishmentsToWingsRepository(
   @Value("\${dpr.lib.redshiftdataapi.athenaworkgroup:workgroupArn}")
   override val athenaWorkgroup: String,
   override val identifiedHelper: IdentifiedHelper,
-) : AthenaReferenceDataRepository<EstablishmentToWing>(athenaClient, tableIdGenerator, athenaWorkgroup, identifiedHelper) {
+  override val userRequestContext: UserRequestContext,
+) : AthenaReferenceDataRepository<EstablishmentToWing>(athenaClient, tableIdGenerator, athenaWorkgroup, identifiedHelper, userRequestContext) {
```
gavriil-g-moj: This would be masking a very sneaky bug if the userRequestContext was used in the repository.
EstablishmentsToWingsRepository is called from a scheduler and is not part of a chained triggered by a request thread so if the repository tried to use it, it would fail.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/AthenaApiRepository.kt
```diff
@@ -60,6 +60,7 @@ class AthenaApiRepository(
   val athenaWorkgroup: String,
   val identifiedHelper: IdentifiedHelper,
   val jdbcTemplate: JdbcTemplate? = null,
+  val userRequestContext: UserRequestContext,
```
gavriil-g-moj: I think we shouldn't couple the repositories with the request context.
See this [comment](https://github.com/ministryofjustice/hmpps-digital-prison-reporting-lib/pull/911/changes#r3189564369)

gavriil-g-moj: Or if we do we need to make this nullable.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/UserRequestContext.kt
gavriil-g-moj: Also apart from error handling as a future note we need to add retries. This was not there earlier and doesn't have to be part of this ticket

gavriil-g-moj:
I left some comments but in summary, the more important points are:
- UserRequestContext needs to be nullable at the Service and Repository layers. Not all calls to these would originate from a request. An example it the scheduled calls to the EstablishmentCodesToWingsCacheService and EstablishmentsToWingsRepository.
- Nullable and ClassCastException case error handling in the UserRequestContext.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/productCollection/ProductCollectionService.kt
ReedSoftware: It doesn't need it, it will always at least return an empty list - it only had it before because authToken also had one and could be null.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/UserRequestContext.kt
ReedSoftware: It should be, but I can certainly do this, better to be safe :)

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/UserRequestContext.kt
ReedSoftware: If you look at the getCaseloads function, it has checks for all this :)

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/establishmentsAndWings/EstablishmentsToWingsRepository.kt
ReedSoftware: Oh that's true, I had completely missed out the scheduled side of things. I think I'd prefer to not make the context nullable, but instead handle the possible edge cases inside of the context if possible. I'll have a think.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/UserRequestContext.kt
ReedSoftware: I've made this class more consistent now - let me know what you think

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/establishmentsAndWings/EstablishmentsToWingsRepository.kt
ReedSoftware: The latest changes should address this?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/AthenaApiRepository.kt
ReedSoftware: The latest changes should address this?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/UserRequestContext.kt
ReedSoftware: Done

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/UserRequestContext.kt
```diff
@@ -0,0 +1,52 @@
...
+
+@Component
+@Scope(value = WebApplicationContext.SCOPE_REQUEST, proxyMode = ScopedProxyMode.TARGET_CLASS)
+class UserRequestContext(
+  private val manageUsersClient: ManageUsersClient,
+) {
+  val authentication: DprSystemAuthAwareAuthenticationToken? by lazy {
+    val securityContext = SecurityContextHolder.getContext()
+    securityContext.authentication?.let {
+      securityContext.authentication as? DprSystemAuthAwareAuthenticationToken
```
gavriil-g-moj: Minor: Here instead of using `it` the `securityContext.authentication` is read again.

gavriil-g-moj:
I'm thinking this doesn't seem the right approach to inject a non nullable request context into the repositories or even the service layer. The boundary for this is the controller. Below that boundary (service, repository) it is not guaranteed a request context will exist. 
As things are now it's not used in the cases where a service or repository is called from a non request context, i.e. a scheduled task, and likely will not (legitimately) be used in these cases as it will not be resolved to anything and if it did try to be used (potentially by mistake as someone could see this parameter is mandatory so would assume they can call it) it would error. Which means it's not right to have it as a mandatory parameter there. 
Overall if we want to take the approach of passing the request context globally I think it needs to be optional in the service and repository layers.

gavriil-g-moj (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/config/ExecutionContextConfig.kt
```diff
@@ -0,0 +1,80 @@
...
+        "/ping",
+        "/v3/api-docs/**",
+        "/swagger-ui/**",
+        "/swagger-ui.html",
+        "/swagger-resources/**",
+      )
+  }
+}
+
+@Configuration
```
gavriil-g-moj: Minor: This is not really configuration. Component would suit better here.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/Scheduler.kt
```diff
@@ -0,0 +1,18 @@
...
+
+interface Scheduler {
+  val scheduler: ThreadPoolTaskScheduler
+  val duration: Duration
+  fun scheduledFunction()
+  fun schedule() {
+    try {
+      scheduler.scheduleAtFixedRate(::scheduledFunction, duration)
+    } finally {
+      ExecutionContext.clear() // Clear this because it's using ThreadLocal, we don't want threads in a thread pool to reuse an existing prepopulated context
```
gavriil-g-moj: I think the fact that we need to know to call clear from schedulers on a static method is a big warning sign. This introduces a hidden dependency.
Regarding the changes here, I think this is not doing what the intention is.
It seems it is clearing the thread local on the thread that calls schedule(), not the thread local used by the scheduled executions.
i.e. the clear should be done on the thread that called scheduledFunction.

gavriil-g-moj:
Wow this is a huge change.
I think there are multiple traps here that we will fall into if we do this. 
I would much rather have the previous implementation and simply have a QueryContext class which holds the username and activeCaseLoadId and passed to the executeQueryAsync. 
This is so much simpler and less risky.
With the current implementation many dangers are introduced:
- Hidden dependencies. The static ExecutionContext class becomes a hidden dependency which makes things less explicit. For example in the scheduler you had to remember to clear the context. Also other places in the code now depend on the ExecutionContext without this being clear.
- Introduction of thread local. I would definitely avoid this. This is a major point of hidden bugs. For example, if an async task is introduced, then this needs to be handled very carefully and it's so easy to go wrong. In the current scheduler, clearing is not performed on the right thread and this is really easy to miss. Also we are using Kotlin. Although we don't seem to use coroutines if someone makes use of this feature, then the request goes to a different thread and the context could either disappear or leak.

gavriil-g-moj (dismissed):

gavriil-g-moj (approved):

PR #936, opened 2026-05-13
DPR2-2800 the status endpoints no longer check if a table exists.

gavriil-g-moj:
The status endpoints no longer check if a table exists. Their responsibility is only to check the status.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 Below you can find the changes included in each release.
 
+# 17.0.0
+- The status endpoints no longer check if a table exists. Their responsibility is only to check the status.
+
 # 16.0.0
 - Extract all user-related data management into an execution context that is populated via an extension method
 - Removed old authentication_code (user token) related code
@@ -15,9 +18,6 @@ Below you can find the changes included in each release.
 # 15.5.0
 - Add an optional `type` field to the DashboardVisualisationColumnDefinition and show this only for the measures.
 This is computed in DashboardDefinitionMapper in a similar way as the ReportDefinitionMapper and not part of the DPD.    
-
-# 16.0.0
-- The status endpoints no longer check if a table exists. Their responsibility is only to check the status. 
 
 # 15.4.1
 - Handle Redshift Data API error response when an execution statement is not found and return 400 instead of 500. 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ Below you can find the changes included in each release.
 - Add an optional `type` field to the DashboardVisualisationColumnDefinition and show this only for the measures.
 This is computed in DashboardDefinitionMapper in a similar way as the ReportDefinitionMapper and not part of the DPD.    
 
+# 16.0.0
```
ReedSoftware: This will need to be 17.0.0 and put at the top

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/DataApiAsyncController.kt
```diff
@@ -266,11 +260,6 @@ class DataApiAsyncController(
       description = "External table ID.",
       example = "reports._6b3c6dfb_f601_4795_8ee5_2ad65b7fb283",
     )
-    @RequestParam(
```
ReedSoftware, line 269: Should we immediately remove? Or deprecate and accept it but just throw it away for now for maybe 1 sprint? Not sure how much impact this would have on other teams who use our API app or integrate the lib.

ReedSoftware (changes requested):

CHANGELOG.md
gavriil-g-moj: Thanks, not sure how this happened.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/controller/DataApiAsyncController.kt
gavriil-g-moj: I have marked it as deprecated. 
This isn't a breaking change as such, as if the user tries to open a report which is not there they will still see the user friendly FE page which will prompt them to rerun the report.

gavriil-g-moj: I have added to the changelog the corresponding FE version which supports these changes for documentation reference.

ReedSoftware (approved):

PR #949, opened 2026-05-20
PDHD-172: schema version

LizDodion:
- Added `SchemaVersionTypeAdapterFactory.kt` to accept the `schema_version` field on the dpds.
  - It treats `v1.0.0`, `1.0.0`, `v1.0`, `1.0`, `v1`, `1` and empty schema_version the same as version `1.0.0`
  - It throws an exception like `Validation failure: Unsupported schema_version: v1.1.0: Only 1.0.0 is supported at this time` if the version is not 1.0.0 (Temporarily until we are ready to start handling different versions)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/SchemaVersionTypeAdapterFactory.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/SchemaVersionTypeAdapterFactory.kt
@@ -1,0 +1,43 @@
+package uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data
+
+import com.google.gson.Gson
+import com.google.gson.JsonElement
+import com.google.gson.TypeAdapter
+import com.google.gson.TypeAdapterFactory
+import com.google.gson.reflect.TypeToken
+import com.google.gson.stream.JsonReader
+import com.google.gson.stream.JsonWriter
+import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.data.model.AnyProductDefinition
+import uk.gov.justice.digital.hmpps.digitalprisonreportinglib.exception.UnsupportedVersionException
+
+class SchemaVersionTypeAdapterFactory : TypeAdapterFactory {
+  override fun <T : Any> create(gson: Gson, type: TypeToken<T>): TypeAdapter<T>? {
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/SchemaVersionTypeAdapterFactory.kt
```diff
@@ -0,0 +1,45 @@
...
+        delegate.write(out, value)
+      }
+
+      override fun read(`in`: JsonReader): T {
+        val jsonElement = elementAdapter.read(`in`)
+        if (jsonElement.isJsonObject) {
+          val jsonObject = jsonElement.asJsonObject
+          if (jsonObject.has("schema_version")) {
+            val version = jsonObject.get("schema_version").asString
+            val validVersions = setOf("1", "1.0", "1.0.0", "v1", "v1.0", "v1.0.0")
```
gavriil-g-moj: All of these are valid but I think we need to define a standard and allow only this versioning standard to avoid inconsistency and complexity. 
Now which version format to define as the standard we allow is up for debate.

src/test/resources/productDefinitionInvalidVersion.json
```diff
@@ -0,0 +1,532 @@
+{
+  "schema_version": "v1.1.0",
```
gavriil-g-moj: To keep it consistent with the rest of the DPD field naming conventions this would be:
`schemaversion` all lowercase one word. 
I know on some occasions the DPD uses camelcase which it shouldn't really. 
The standard for the fields is:
- all lowercase
- plural as singular i.e. report instead of reports

LizDodion: cool. this was a suggestion from des but happy to change :) 

This is useful info

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/SchemaVersionTypeAdapterFactory.kt
LizDodion: i'm happy for 1.0.0 if we're all happy with it :)

ReedSoftware (dismissed):

ReedSoftware (approved):

PR #950, opened 2026-05-21
check authsource against given env vars for allowed auth sources

ReedSoftware:


gavriil-g-moj:
Can we also update the changelog please?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/security/DprSystemAuthAwareTokenConverter.kt
```diff
@@ -1,21 +1,32 @@
...
+          jwt,
+          authAwareAuthenticationToken.clientId,
+          authAwareAuthenticationToken.userName,
+          authAwareAuthenticationToken.authSource,
+          extractAuthorities(jwt, JwtGrantedAuthoritiesConverter()),
+        )
+      }
+
+    if (!requiredAuthSources.map { it.lowercase() }.contains(token.authSource.name.lowercase())) {
+      throw InsufficientAuthenticationException("You are not authorised to view this application.")
```
gavriil-g-moj: Will this result in 401?
Isn't this more like a 403?
Would the AccessDeniedException be more suited here?
https://docs.spring.io/spring-security/site/docs/4.2.19.RELEASE/apidocs/org/springframework/security/access/AccessDeniedException.html

gavriil-g-moj: Because the user is authenticated but lacks the required authority source to view the application.
What are your thoughts? 
Maybe nit picking.

ReedSoftware: Ah you're right, this is incorrect _authorisation_ not _authentication_ - the user may well have authenticated, just with the wrong type of user. Will fix.

src/test/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/integration/AuthSourceIntegrationTest.kt
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+class EmptyAuthSourceEnvVarAuthSourceIntegrationTest : IntegrationTestBase() {
+  companion object {
+    @JvmStatic
+    @DynamicPropertySource
+    fun setupClass(registry: DynamicPropertyRegistry) {
+      registry.add("dpr.lib.definition.locations") { "productDefinition.json" }
+      registry.add("dpr.lib.user.requiredAuthSources") { "" }
```
gavriil-g-moj, line 39: What will happen if the requiredAuthSources is not set at all?
Will it be null? 
And what will the behaviour be in this case?

ReedSoftware: You can't "unset" a property so I can't add a test case here, but the value code is
`@Value("#{'\${dpr.lib.user.requiredAuthSources:nomis}'.empty ? 'nomis' : '\${dpr.lib.user.requiredAuthSources:nomis}' }") private val requiredAuthSources: String`

Which means that it will default to `nomis`

gavriil-g-moj (approved):

PR #973, opened 2026-06-26
Pdhd 211 lao policy

ReedSoftware:


src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/policyengine/Policy.kt
```diff
@@ -22,15 +23,44 @@ data class Policy(val id: String, val type: PolicyType, @SerializedName("action"
       }
       break
     }
+
+    if (type == PolicyType.LAO) {
+      if (rule.size != 1) {
+        throw IllegalStateException("LAO policy without rule")
+      }
+      if (rule.first().effect != Effect.PERMIT && (_action == null || _action.size != 1)) {
```
gavriil-g-moj, line 32: Here if the rule is not "permit" we throw an exception with the message:
"LAO policy provided without accompanying CRN column".
Shall we have a separate case for this with a separate message instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/policyengine/Policy.kt
```diff
@@ -22,15 +23,44 @@ data class Policy(val id: String, val type: PolicyType, @SerializedName("action"
       }
       break
     }
+
+    if (type == PolicyType.LAO) {
+      if (rule.size != 1) {
+        throw IllegalStateException("LAO policy without rule")
```
gavriil-g-moj: Since we don't handle IllegalStateException in `DigitalPrisonReportingExceptionHandler` this will result in a 500 response. 
It's better to add an `@ExceptionHandler` in `DigitalPrisonReportingExceptionHandler` and return a 400 in this case. 
Also if we don't want to globally return 400 for IllegalStateException, we can throw a custom exception and handle that instead.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/service/PolicyEngine.kt
```diff
@@ -23,10 +23,12 @@ class PolicyEngine(
 
   fun execute(): String = doExecute(policy.sortedByDescending { it.type })
 
-  private fun doExecute(policiesToCheck: List<Policy>): String = if (policiesToCheck.isEmpty() || isAnyPolicyDenied(policiesToCheck)) {
+  fun checkLaoPolicyExists(policies: List<Policy>): Boolean = !executionContext.hasProbationDatasources || policies.any { it.type == PolicyType.LAO }
```
gavriil-g-moj: Minor: Optionally for readability maybe `checkLaoPolicyExists` could be renamed to `laoPolicyExists`. 
As it's doing a double negation (!executionContext.hasProbationDatasources and later !checkLaoPolicyExists(policy))

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/policyengine/Policy.kt
```diff
@@ -22,15 +23,44 @@ data class Policy(val id: String, val type: PolicyType, @SerializedName("action"
       }
       break
     }
+
+    if (type == PolicyType.LAO) {
+      if (rule.size != 1) {
```
gavriil-g-moj: In the case when someone defines more than 1 rule then the message: ""LAO policy without rule" is misleading.

src/main/kotlin/uk/gov/justice/digital/hmpps/digitalprisonreportinglib/data/model/policyengine/Policy.kt
ReedSoftware: Im a bit confused - the only time we throw an exception here is if rule is not permit AND there's no action.

gavriil-g-moj: Yes true, I've resolved this comment.

gavriil-g-moj (approved):

PR #981, opened 2026-07-07
temporary solution to stop access to probation data from non probation context

ReedSoftware:


gavriil-g-moj (approved):

gavriil-g-moj:
Agreed this is a temporary solution and will be fixed properly later.

PR #982, opened 2026-07-07
protect single dashboard variant endpoint with policy auth check

ReedSoftware:


gavriil-g-moj (dismissed):
Approved but some existing tests now fail since probably these dashboards would be unauthorised now.

gavriil-g-moj (approved):

